PR #95, opened 2026-08-09
workflow(aw): sanction single-context Full run when sub-agent dispatch is unavailable

mthines:
In harnesses that disable the Task tool (e.g. Claude Code on the web), the aw
dispatcher cannot hand a Full-tier task off to aw-planner -> aw-executor. The old
fallback was under-specified and self-contradictory: the dispatch section said
"invoke them yourself if your tools permit" while the Hard rules forbade aw from
ever using Edit/Write on a Full task, so an aw run in the cloud env hit
"Failed to run agent" and had to improvise a justification before proceeding.

Make the emergent behavior the sanctioned, reproducible path: when the split is
structurally unavailable, aw runs the Full phases in one window -- planner role
(Phases 0-2, plan.md + checks.yaml, clearing confidence(plan) >= 90%) then
executor role (Phases 3-7). This preserves the plan artifact and the confidence
gate (the two load-bearing wins of the split) and concedes only context
isolation, which the harness has already made impossible. It is explicitly not a
downgrade to Lite/Micro single-pass, and never a license to skip an available
split. The "tell the user to run the agents themselves" text is demoted to a
last resort for when Edit/Write/Bash are also unavailable.

Four files change:

- aw.agent.md: rewrite the Full-tier dispatch fallback; condition the
  Edit/Write/Bash Hard rule on dispatch availability; condition the Routing
  table's Full row the same way, so the summary table no longer states an
  unconditional "dispatch only" / "Never use Edit/Write/Bash in this tier".
- routing.rule.md: matching outer-layer fallback -- when the main session cannot
  dispatch aw, load the workflow and play the dispatcher role in-context -- plus
  a pointer to the aw agent definition for the step-by-step procedure.
- SKILL.md: version 3.18.0 -> 3.19.0, and the Templates section now states the
  single-context Full fallback and links templates/aw.agent.md for the
  procedure. Without this, a reader following the routing rule's own
  "Skill(autonomous-workflow)" instruction landed on a document that described
  the split as the only Full route.
- CLAUDE.md: update the thin-router invariant, the split-is-Full-only
  design-intent invariant, the smoke test (step 3), and add a v3.19.0 history
  entry enumerating all four coupled surfaces.

Tier-detection table unchanged (L1 Check B keeps it byte-identical to SKILL.md
Step 1); checks.yaml / confidence(plan) integrity rules unchanged. L1: 171/171.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01HFbsywyxg9JDS53F5Nvwbw

### DIFF
--- a/skills/workflow/autonomous-workflow/CLAUDE.md
+++ b/skills/workflow/autonomous-workflow/CLAUDE.md
@@ -393,7 +393,12 @@ When editing this skill, do not break these — they're load-bearing:
 - **`aw` is a thin, opt-in router.** It detects the tier and routes (single-pass
   for Micro/Lite, planner→executor for Full) and owns the universal lessons loop.
   It must not accrue domain knowledge, must not force Full on light tasks, and
-  must not silently wrap casual edits. The planner/executor split is Full-only.
+  must not silently wrap casual edits. The planner/executor split is Full-only —
+  **except** when the harness disables sub-agent dispatch (`Task`), where `aw`
+  runs the Full phases single-context (planner + executor roles in one window)
+  rather than downgrading to Lite. That fallback preserves the `plan.md` artifact
+  and the `confidence(plan)` gate; it is the one place `aw` uses `Edit`/`Write` on
+  a Full task, and it must never be used to sidestep an *available* split.
 - **Episodic lessons are advisory-only.** An `aw-lessons` lesson biases the plan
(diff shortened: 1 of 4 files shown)


mthines:
@dash0 implement

mthines:
@dash0 implement

PR #103, opened 2026-08-12
fix(reviewer): forbid coordinate keys in comment-relevance memory

mthines:
## Why

The `reviewer-comment-relevance` bucket is meant to key on a structural `<category>:<claim-gist>` fingerprint so the signal **accumulates** across files, commits, and PRs. In practice the `pr-reviewer` write path drifted into per-comment **coordinate keys** (`pr{N}-{commentId}-{slug}`), which:

- created 2–3 duplicate rows per comment (the slug is re-derived each run), and
- **defeated accumulation entirely** — a key unique per comment never reaches `seen_count ≥ 3`, so suppression never fires and the memory goes inert.

Found while grooming `repo::dash0hq/dash0`: ~40 redundant rows across 36 comments, several carrying contradictory `fixed` vs `wont-fix` verdicts for the *same* comment. The old wording (*"keys MUST NOT encode `file:line`"*) was read too narrowly — the agent inferred a PR/comment-id coordinate was fine.

## What

- **`comment-relevance-memory.md`** — broaden the prohibition to **any** coordinate (PR#, comment-id, SHA, thread-id), add ✅/❌ examples of the exact observed anti-pattern, a mechanical-derivation instruction, and a pre-write self-check.
- **`thread-resolution.md`** — mirror the warning in the pr-reviewer write path with a cross-link.
- **`implement-suggestion/SKILL.md`** — expand the Phase 7 emit's key template to `<category>:<claim-gist>` so the ban covers every write path, not just `pr-reviewer`'s.
- **Derivation** — the rule now names `fingerprint()` normative and spells out its four steps (positional first-6 words, `suggestion` default when no prefix, `nit` → `nitpick`) instead of describing a summarisation the script does not perform.
- **Migration** — an explicit note that pre-existing bare-gist rows are left to lapse on TTL rather than bulk re-keyed.
- **`scripts/eval/l1.mjs`** — 8 new `G9d` guards locking the contract (5 static, plus one per writing agent asserting no bare `<fingerprint>` placeholder survives).

## Verification

`node scripts/eval/l1.mjs` → **178/178 checks passed**. Each new guard was proven red against the pre-fix content before being accepted (174/175 and 176/178 respectively). No behavior change to any script; `record-comment-relevance.mjs` already produced the correct fingerprint — this aligns the agent-side rules with it.

Review round 2 (4 threads, all resolved): tightened the self-check thresholds the first round introduced — row 3 raised to `\d{9,}` (a `120000` ms timeout and a `1048576` byte cap are ordinary claim content), and row 4 tightened to require a hex letter as well as a digit, which was matching any 7-digit number for the same reason. G9d now bans both the `<fingerprint>` and `{fingerprint}` placeholder spellings.

Review round 1 (3 threads, all resolved): narrowed the pre-write self-check so it keys on coordinate *shapes* rather than any 3+ digit run — `fingerprint()` preserves digits, so the old test tripped on legitimate gists like `issue:500-responses-not-retried` and would have made the agent strip digits the script keeps.

The 41 already-accumulated duplicate rows on `dash0hq/dash0` were archived out-of-band (reversible) during the groom; this PR stops the recurrence.

### DIFF
--- a/agents/shared/rules/comment-relevance-memory.md
+++ b/agents/shared/rules/comment-relevance-memory.md
@@ -147,6 +147,17 @@ magnitude in words (`four-gib-allocation-cap`, not `4294967296-byte-cap`), which
 is the better gist anyway. If you have encoded a coordinate, move it to the
 record's `examples` field.
 
+**Existing rows are not migrated.** Most rows written before this rule carry a
+bare gist with no `<category>:` segment (at the time of writing, 59 of 70 rows in
+`repo::mthines/agent-skills`). Re-keying them in place is deliberately **not**
+attempted: a bulk rewrite would have to re-derive `category` from comment bodies
+that are no longer fetched, and a wrong guess is worse than no row. Instead the
+old keys are left to lapse — they are never re-sighted under the new derivation,
+so their TTL expires them, while the correctly-keyed row for the same finding
+class accumulates from `seen_count: 1`. Expect a one-TTL window in which a
+finding class is represented by both an old and a new row; suppression simply
+takes longer to arm for those classes and nothing is lost.
+
 ---
 
 ## Relevance memory record schema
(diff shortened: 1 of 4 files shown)


mthines:
@dash0 implement

PR #107, opened 2026-08-15
Bound every external wait so `aw` runs stop hanging on Claude Cloud

mthines:
## Why

`aw` runs on Claude Code for web/mobile go silent for long stretches and never terminate until the user interrupts.

Investigating turned up three things the workflow assumes about the machine it runs on. All three are false in the cloud, and each produces a hang on its own:

- **`gh` and `gw` are absent** (verified `which gh` → exit 1). 169 call sites across the aw rules, `create-pr`, `review-loop`, `implement-suggestion`, `ci-auto-fix` and the agents have no MCP mapping, so every failure becomes an ad-hoc retry-or-improvise decision.
- **`timeout 1800` can never fire.** Claude Code caps a Bash call at 600 s, so the harness kills the watch before `exit 124` — the documented expiry handling was dead code, and the documented recovery was to *re-watch*, uncounted.
- **`pr-reviewer` is `Task`-only with no fallback**, so `review-loop` failed as a mid-Phase-6 tool error the caller had to interpret.

## What changed

Only the verified subset. The full diagnosis lives in `references/cloud-stalling-analysis.md`, which also carries an adversarial review of its own remediation plan — that pass **falsified the premise** behind the capability probe and degradation matrix (the `Task` tool *is* available in the probed cloud session), so those were deliberately not built.

- Every wait is bounded and fits under the harness ceiling: `timeout 540` per attempt against one PR-scoped `CI_WATCH_ATTEMPTS` budget (~36 min total — *more* real wait than the broken `1800`). Phase 7 had no bound at all; it now shares that budget and skips when `create-pr` already drove the checks to a terminal state.
- `review-loop` checks for sub-agent dispatch in Step 0 and returns a logged deviation. Deliberately *not* an in-context substitute — `pr-reviewer`'s independence comes from the isolated context, so playing the role would be a self-review wearing a reviewer's label.
- `aw` gained an `AW RUN COMPLETE` terminal block. For a sub-agent the final message *is* the return value, so a run could previously end with nothing — indistinguishable from a hang. The `Degraded:` line is mandatory so a skipped review is reported rather than a PR being described as reviewed.

Only the CI watch was deduplicated. The two `review-loop` passes are **not** duplication — Phase 6 reviews the pre-CI diff, Phase 7 reviews after `ci-auto-fix` may have pushed. An earlier draft of the analysis called them redundant; that was wrong and is corrected in the doc.

## How to verify

- `node scripts/eval/l1.mjs` → 204/204
- The remediation table in `cloud-stalling-analysis.md` records what landed and what did not, with the finding that blocked each one

## Still open

The `gh` shim (the steelman that beat the capability probe), and one real transcript of a stalled cloud run — the falsified premise means the `Task` branch of the diagnosis may be treating a phantom.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MinNF41BERzCMVu3HdbLtx)_


mthines:
## Disposition of the 27 review threads

These were not read before merge. `create-pr`'s Step 6.7 external-bot feedback loop was skipped when `gh` turned out to be absent from the environment, and no MCP substitute was run — so only the `pr-reviewer` subagent's rounds were consumed while `dash0-dev` was reviewing in parallel. That is a process failure, not a disagreement with the findings: **every live one was correct.**

Follow-up: **#109**.

### Fixed during #107's own later rounds (threads now outdated)

| Finding | Landed in |
| --- | --- |
| `timeout 540` still dies at the 120 s tool default | `e4c5bf4` — explicit `timeout: 600000` at all sites |
| A shell variable cannot carry the budget across the Phase 6/7 boundary | `e4c5bf4` — file-backed `.agent/ci-watch-<pr>.state` |
| No serialization format for the state keys | `e4c5bf4` — `key=value` lines + `get`/`put` |
| `CI_WATCH_ATTEMPTS` named in `diagnostic-surface.md` | `e4c5bf4` |
| Registration probes spending the completion budget | `46cbf3d` — split into `registration_attempts` |
| Immediate `no checks reported` retries finish in seconds | `46cbf3d` — bounded 90 s poll |
| `gh pr checks` exits 8 while pending → tooling arm fires | `e3e9655` — polarity inverted, breaks on pending |
| Confirming poll's status leaks as the script's | `e3e9655` — explicit `exit 0` |
| `declaresToolTimeout` matched anywhere in the file | `e3e9655` — per-site proximity window |
| `isWatch` treating any `bash -c` as a wait site | `e3e9655` — requires a loop keyword + `sleep` |

### Live on merged `main` — fixed in #109

| Finding | Fix |
| --- | --- |
| `create-pr` Step 8 "spends from the same file" vs `ci-auto-fix`'s informational-only rule | Step 8 restated to the read-only / report-back contract |
| `phase-7` claiming "the same rule as Step 9" while stating a different one | `phase-7` now defers to `create-pr` as the owning surface |
| `attempts` compared before increment → 5 watches, not 4 | Increment point made explicit |
| `registration_attempts` same off-by-one → 4 polls, not 3 | Same |
| "debit `attempts`" — wrong direction, the counter rises toward the cap | Now adds, in both surfaces |
| `triaged failure` — a value nothing writes; `failing` matched no row | Rows key on `green` / `failing` / `pending`; `failing` routes to Auto Fix |
| `outcome` enum mismatch with what `ci-auto-fix` returns | Uses `ci-auto-fix`'s vocabulary verbatim |
| No exit-127 tooling row in `phase-7` | Mirrors `create-pr` Step 7 |
| `Degraded:` dropped from `aw`'s Micro/Lite one-liner | Example now carries it |
| `CI_WATCH_ATTEMPTS` in `CLAUDE.md` history + the analysis doc | Removed; zero hits remain |
| `get`/`put` are functions that don't survive a fresh Bash call | Labelled a per-call snippet — **this one my triage also missed**, caught only on a second pass over these threads |

### Answering the one question

> How should the agent confirm `Task` is available without dispatching?

Honestly: **it can't, reliably.** There is no capability-introspection API, and a refused dispatch may surface as an uncatchable harness error rather than a return value. The Step 0 precondition is therefore best-effort — it catches the case where the agent can see `Task` is absent from its toolset, and degrades to the error path otherwise. The reason it is still worth having is placement, not certainty: a failure at Step 0 produces one clean logged deviation, where the same failure at Step 1 produced a mid-Phase-6 error the caller had to interpret. Worth revisiting if capability introspection ever exists.

Resolving these threads now; #109 carries the code.

---

PR #120, opened 2026-08-16
fix(pr-reviewer): make the report format reachable at the point of posting

mthines:
## Summary

`lorekit#482` and `lorekit#492` both posted flat, marker-less reports — gate table dumped straight into the review body, no `Review details` accordion, no `<!-- PR_REVIEWER_REPORT -->`. `#492` ran **ten hours after #119 merged**, on a runner that fetches `agent-skills` at run time.

So the templates were correct and were not being followed. This PR fixes *why* they weren't reachable at the moment the body gets composed.

## The three causes

**1. The template section was named `### Review body format`.** An agent building a review body looked that up, found the report templates, and posted the report as the review body — `F-report-in-review-body`, handed to it by the heading. The section had to disclaim itself in its own opening sentence (*"This is `REPORT_BODY` — the body of the sticky comment, **not** the review's pointer body"*), which was the tell.

**2. Step 4a sits ~280 lines above the templates** in a 2700-line file, so the format is out of view when the body is built.

**3. Step 4a never said "render verbatim."** Grepping its 90 lines for *verbatim* / *literal* / *template* returned one incidental hit inside a table cell. The observed drift is exactly what that predicts:

| Observed in the wild | Spec says |
|---|---|
| `cleared review, no anchor in the diff` | `cleared review, not inlined` |
| ``Reviewed `8a1d2ec` —`` opening a full report | that's the Step 4b *pointer* opening |
| `**Verdict**`, `**CI**`, `**Prior findings**` sections | not in any template; the advisory verdict is terminal-only |
| `<sup>Reviewed for commit…</sup>` at top level | belongs inside the accordion |

`#492` also rendered `<details>` correctly for the *optional* `Additional findings` block while omitting the *mandatory* accordion — so this was never a rendering-capability problem.

## Changes

- **Rename** `### Review body format` → `### REPORT_BODY format (the sticky comment)`, sweeping every call site including `reviewer-report-ingest.md`. The section now records why the old name is not coming back.
- **Verbatim contract at Step 4a** — a table of the five observed drift shapes and what each produces, placed where the body is actually composed rather than 280 lines away.
- **`report_body_is_safe()`** — a Step 4a pre-flight mirroring the existing `payload_is_safe`: marker present, `Review details` accordion present, no `open` attribute, nothing the accordion owns rendered above it, no `**Verdict**` line. A failing body is re-rendered once; if it still fails the run reports verbatim and posts **no report** — the sticky is left as it was, while the separate Step 4b review object still carries any inline findings.

This is **the first check on what is actually posted**. Every prior guard verifies the spec file, which by construction cannot catch a run that reads it correctly and writes something else.

- New `F-report-body-composed-from-memory` failure mode + the pre-flight added to the hard-invariants list.

## Verification

`node scripts/eval/l1.mjs` → **324/324**.

New **G24p** locks all three fixes. It asserts the pre-flight is **not a stub** — each claimed assertion must actually be present in the body — and that the verbatim contract sits between Step 4a and the templates **by position**, not merely by existing somewhere.

Mutations verified red:

| Mutation | Result |
|---|---|
| Revert heading to `Review body format` | exit 1 |
| Move verbatim contract after the templates | exit 1, 322/324 |
| Delete the verbatim contract | exit 1 |
| Stub `report_body_is_safe` to `return (True, "")` | exit 1 |
| Drop only the accordion assertion from the pre-flight | exit 1 |
| Unregister the failure mode | exit 1 |

Confirmed no false positive on the `**Verdict**` rule: it appears only in the Step 3 *terminal* blocks, in zero of the three `REPORT_BODY` templates.

## Note

Two mutations (reverted heading, deleted contract) fail via a `sliceBetween` crash rather than a named check. Still a hard non-zero exit, just less informative output.

---


mthines:
@dash0 implement

PR #121, opened 2026-08-17
refactor(pr-reviewer): render the report from one template, not from prose

mthines:
## Summary

Five production runs — `mthines/lorekit#482`, `#492` x3, `#495` — posted reports with **no `<!-- PR_REVIEWER_REPORT -->` marker and no `Review details` accordion**, from a spec that mandated both. The last ran **ten hours after #120 merged**, on a runner that fetches `agent-skills` at run time. So this was never staleness and never a capability limit: the layout was simply unfollowable.

This stops asking the model to render layout.

## Why the previous PRs didn't fix it

Measured, not guessed:

| | before |
|---|---|
| Report templates | 3, **71-86% identical** to each other |
| Distance from Step 4a to the templates | 280 lines |
| Rule bullets in the format section | 16 |
| Distinct `ALL_CAPS` placeholders in the file | 124 |

Runs weren't copying three near-duplicate templates — they were averaging them into a remembered shape. And **four of the five report-rendering failure codes were mine**, added in #119 and #120. Each round answered drift with more prose, when the volume of prose was the cause.

## Changes

**Layout has exactly one home.** `agents/pr-reviewer/templates/report-body.md` — one template with conditional `{{#BLOCK}}` sections, replacing all three variants. An absent optional key removes its block, so "omit the placeholder when empty" becomes the mechanism rather than a rule to remember.

**Filling it is deterministic.** `agents/pr-reviewer/scripts/render-report.mjs` takes a JSON payload and **fails closed** on: unknown key, missing required slot, incomplete companion group, orphaned dependent slot, invalid gate glyph, smuggled `**Verdict**` line, a markdown link caged in a code span, unresolved placeholder, or a template that lost its marker/accordion. Every failure exits non-zero and prints **nothing** on stdout, so a piping caller cannot post a fragment.

**Step 4a builds data and shells out.** Hand-rendering as a fallback is explicitly forbidden: on renderer failure, report the error and leave the sticky untouched. Path resolution uses a portable symlink walk (`readlink -f` is GNU-only) and aborts distinguishably when it cannot locate the renderer.

**A pre-write net that survives a bypass.** Four assertions run on `REPORT_BODY` immediately before the write, whatever produced it — the renderer's own post-conditions only fire if the renderer fires. The accordion check is POSIX `awk`, because both plausible `grep` forms are wrong: the plain-quoted pattern treats the escape as the letter n, and the ANSI-C form makes the newline a pattern *separator*, i.e. an OR that passes on either half.

[...]

PR #122, opened 2026-08-17
refactor(pr-reviewer): make the report payload data, not markdown

mthines:
Follow-up to #121, which deliberately deferred this.

## Summary

#121 made the *layout* deterministic and I called the report consistent. Consistency has two halves — structure and content shape — and only the first was done. Six value-shape defects were visible in the very first report the new pipeline posted:

| Defect | Consequence |
|---|---|
| `FOOTER_LINE` carried a 40-char sha while `CI_NOTE` carried 7 | two sha forms in one report |
| `RUN_MODE` read `750 additions / 486 deletions` | the contract says `<N> lines in delta`, and `reviewer-report-ingest.md:114` parses that slot for `delta_lines` — **the first report the pipeline produced was unparseable by its own documented grammar** |
| No cap on gate cells | a 400-char or multi-line value silently breaks the table |
| A markdown link caged in a code span | shipped a dead link on `#121`'s own sticky |
| Orphaned `RESOLVED_SINCE` | dropped silently at exit 0 (fixed in #121) |
| Stray nested field inside `RUN` | accepted in silence |

Every one is a value the model wrote as markdown or as a pre-formatted string. So the payload becomes **data**.

## Changes

**Counts are never supplied.** `Open bot threads (N)`, the `<summary>` suffix with its singular/plural and ` (K blocking)`, and the three findings counts all come from array length. A count cannot disagree with its list, and a list cannot be orphaned from its counter — so the companion `GROUPS` mechanism and the one-way dependency check from #121 are both **deleted**, because the failures they guarded are now unrepresentable rather than rejected.

**Links are never supplied.** `OPEN_THREADS`, `ADDITIONAL_FINDINGS`, `LOW_CONFIDENCE_FINDINGS` and `MEMORIES_USED` take `{path, line, url?, …}` and the renderer builds the markdown. Link syntax inside those fields is rejected, so the caged-link class cannot recur. Backticks are rejected in the identifier fields the renderer wraps itself (`path`, `key`) and allowed in the prose ones (`ask`, `body`, `note`), because `ask` reproduces another bot lead line verbatim and those name symbols in code spans. A missing `url` renders unlinked inline code, never a broken link.

**Shapes are derived.** A `RUN` object `{mode, sha, prior_sha, delta_lines}` produces both the footer line and the `<mode> · <N> lines in delta` line — one shape for every mode, a zero-delta run included, so the grammar always has a `delta_lines` to parse. The `<N> used` half of the memories line is derived from `MEMORIES_USED` too. A sha must be exactly 7 lowercase hex chars. Extra run colour goes in `RUN_NOTE`, appended *after* the parseable prefix — that's the `CI_NOTE` lesson from #121: a spec that forbids what the run must say gets ignored, so give it a legal home.

**Everything else validated:** gate cells single-line, no pipe, ≤120 chars; log slots must begin `ran`/`skipped`; `QUALITY` must begin `produced`; nested objects get the same unknown-key discipline as the top level. A v1-shaped payload fails loudly and the error names the replacement.

`OPTIMALITY_CARDS` stays model-authored markdown — a card is a multi-line block with its own table. Each must carry its `### Optimality proposal — <path>:<line>` heading, which the ingest grammar keys on and the renderer checks.

## The missing test

**A producer → consumer round trip.** The extractors documented in `reviewer-report-ingest.md` are applied to real rendered output, and every section must parse to its documented type — Headline, Footer SHA, Run mode, Standards log, Optimality log, Skipped files, the gate table as five typed rows, plus the conditional sections where present. Derived counts are additionally checked to equal the bullets rendered under them.

This is what was missing when `Run mode` shipped unparseable. **An unenforced grammar is aspirational** — which is exactly the failure this line of work has been diagnosing. I fixed it one level up in #121 and reproduced it one level down.

Writing the Footer extractor also surfaced that the grammar's own instruction matters: it says match on `` commit `<sha>` `` *alone, never on a leading phrase*, and my first attempt anchored on the phrase and matched only the incremental form.

## Verification

`node scripts/eval/l1.mjs` → **496/496** (from 356).

Every derivation and validator mutation-tested red. **Four were initially unguarded** — the 7-char sha, the 120-char cap, stray nested fields, and markdown-in-structured-fields all stayed green when disabled, because I'd written the validator without a test. Now covered by the G25 `rejects` table, which additionally asserts that no case is passing merely because its slot became an unknown key. One apparent green turned out to be my `perl` escaping failing to apply rather than a weak guard, confirmed by re-running the mutation through Python.

Snapshots regenerated; the three `.expected.md` files remain the human-readable reference.

---


mthines:
@dash0 implement

PR #123, opened 2026-08-18
feat(review-loop): converge CI, and support an out-of-process reviewer

mthines:
## Why

`review-loop` converged review threads but had no CI awareness at all — even though `pr-reviewer`'s Gate 2 **fails the verdict on a red build**. Its convergence test was `unresolved_thread_count == 0` only, so a standalone `/polish` could burn five iterations, push on every one, and exit `FAIL`-because-CI having never read a check run.

The question behind this change was whether `implement-suggestion` should fix CI. It should not: its worker hard-stops a failing batch and pushes nothing, while `ci-auto-fix` pushes-watches-reverts. Those are opposed push disciplines, and their gates are calibrated differently on purpose. The composition point already existed — `review-loop` — so that is where CI landed.

## What changed

**`review-loop` sub-step D (CI).** After each iteration's push, read check state statelessly (`gh pr checks`, never a watch) and delegate a red mechanical failure to `ci-auto-fix`, capped at 2 handoffs per run. Convergence now means zero open threads **and** CI not red. `--no-ci` opts out.

**`review-loop --external-review`.** Sub-step A waits for a reviewer running in another process (an automation-triggered agent, a review bot) instead of dispatching `pr-reviewer`. Sub-steps B/C/D are unchanged — they read threads from GitHub and do not care who authored them. Because it dispatches no `pr-reviewer`, it is also the graceful-degradation path on harnesses where `Task` is disabled, which previously forced a hard skip. `--interval S` (clamped 540) sets the poll bound; `--no-feedback` is refused and `--critical` warns-and-ignores.

**`implement-suggestion` keeps its boundary, with a tighter pre-push gate.** Its worker now runs one full **unscoped** fast-check pass before the single push (step 3.5). The per-comment checks are scoped to touched files, so a rename could commit clean per-comment and still leave a consumer broken elsewhere — the "small type error" case that turns CI red for a reason the worker already had in hand. Under `--watch`, post-push check state is read only as a **stop reason** (`ci red — <checks>`); no dispatch, no budget spent.

## One owner of the CI budget per invocation

This is why `--no-ci` exists rather than the CI step being unconditional. Watch state is queried, never carried (`diagnostic-surface.md`), so nothing coordinates the budget through a shared file — each skill counts its own handoffs inside its own invocation, and exactly one of them owns CI over a given stretch of the PR:

- **`create-pr` Step 6.5** passes `--no-ci` — Steps 7–9 run afterwards and already cover every commit the loop pushes.
- **`autonomous-workflow` Phase 7** passes `--no-ci` **plus a new Step 2.5**. That phase invoked `review-loop` after CI went green and then handed back with no re-watch — so the loop's own commits landed on a head its watch never saw. Step 2.5 re-checks CI once the loop has pushed, spending that phase's existing `attempts` budget rather than opening a second one.
- **`review-changes`** passes nothing and gets CI convergence, since it owns no CI phase.

Sub-step D is a query plus a delegation, so it adds no `gh … --watch` site and records no verdict or spent budget for another phase to inherit.

## Shared poll, not a copy

The hardened review-activity probe moves to `agents/shared/rules/review-activity-poll.md` — one owner, two callers — instead of being duplicated. It encodes five properties that are individually easy to drop and collectively load-bearing:

- `updated_at` over `created_at`, so a sticky report **rewritten in place** counts as new feedback
- `--paginate` with a line-per-match `--jq`, because the REST default page size is 30 and `gh` applies `--jq` once per page
- a guarded `count()` separating a broken `gh` from a legitimately empty result
- the 540 s interval clamp beneath the 600 s Bash tool cap
- the bare-fence requirement (the block contains apostrophes; a wrapping `bash -c '…'` silently breaks it)

`POLL_ERROR` aborts in both callers and is never read as "reviewer quiet" — converting a broken probe into silence would report a never-reviewed PR as converged, with a stop reason that reads like success.

## Verification

- `node scripts/eval/l1.mjs` → **496/496**, unchanged from the pre-change baseline.
- The poll move is **G22/G23 site-neutral by design**: the block moved between two walked directories and was not duplicated, so `EXPECTED_SITES = 9` and the 3-polling-block pin both hold without edits to `l1.mjs`.
- Every `Skill("review-loop", …)` call site was swept, not just the two that needed the flag.

L1 is a contract checker, not a behavioral test — these are instruction files, so the real validation is a reviewer reading the boundaries.

## Reviewer attention

1. **Step 2.5 in `phase-7-ci-gate.md`** is the one place I widened scope beyond the stated plan. Passing `--no-ci` there alone would have *preserved* an existing gap rather than closing it, so I closed it with that phase's own budget. Worth confirming that is the division you want.
2. **`--external-review` iteration-1 semantics**: a `quiet` poll still runs one pass, because the external reviewer may have reviewed before the loop started. Without it, invoking the loop on an already-reviewed PR would exit immediately having done nothing.
3. **The `"unread"` arm of `ci_is_settled()`** covers a PR that arrives already reviewed and thread-clean, where iteration 1 reaches the convergence exit before sub-step D has ever run.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ebss85eNMoo1ErRAxxFDXB)_


mthines:
@dash0 implement

PR #124, opened 2026-08-18
feat(pr-reviewer): validate the POSTED report shape from outside the agent

mthines:
## Summary

`lorekit#503` showed the remaining failure is **not** that the spec is unfollowable — #121 and #122 fixed that. It's that following it is **optional**. One definition, one PR, one head:

| Time (UTC) | Review | Shape |
|---|---|---|
| 17:44 | `4964076700` | correct — accordion, derived counts, real links |
| 17:56 | `4964171425` | correct |
| 18:07 | `4964277130` | correct |
| **18:31** | `4964475125` | **flat** — no marker, no accordion, gate table at top level |

Both merges predate all four runs by ~32 hours, so this is neither staleness nor a spec defect. It's non-determinism — **3 of 4**, up from 0 of 7 before this work — and nothing catches the one that drifts.

## Why nothing caught it

Every guard built so far runs where the agent decides:

| Guard | Where it runs | Sees a posted report? |
|---|---|---|
| L1 / `G25` (439 checks) | CI, against the repo | **No** — it validates the repository |
| `render-report.mjs` fail-closed | inside the run | only when it is invoked |
| Step 4a pre-write assertions | inside the run | only when the run reaches them |
| ingest round-trip | CI, against fixtures | **No** — committed fixtures, not live output |

The flat body has neither marker nor accordion, so pre-write assertions 1 and 2 would *both* have aborted — had they run. **Four PRs made the correct path easier and better-guarded, and never added a check that sees reality.** This adds one.

## Changes

**`scripts/validate-report-shape.mjs`** — takes a posted body, emits a structured verdict:

| Code | Catches |
|---|---|
| `missing-report-marker` | report sections with no `<!-- PR_REVIEWER_REPORT -->` |
| `report-marked-as-pointer` | a full report stamped `<!-- PR_REVIEWER_POINTER -->`, so a later run recovers it as a pointer and reads the report as absent |
| `no-review-details-accordion` | the flat regression |
| `accordion-pre-expanded` | a `<details>` carrying `open` |
| `accordion-owned-line-at-top-level` | gate table / run mode / memories / footer above the accordion |
| `verdict-in-posted-body` | the terminal-only advisory verdict |
| `link-caged-in-code-span` | the dead-link defect from #121 |
| `count-disagrees-with-list` · `summary-count-disagrees` | a declared count ≠ the bullets under it |
| `pointer-too-long` · `pointer-carries-sections` | a pointer grown into a report |

A body that is neither report nor pointer is ignored, so ordinary PR comments never trigger it.

**`.github/workflows/reviewer-report-shape.yml`** — reusable workflow on `pull_request_review` + `issue_comment`. Posts **one sticky notice** per PR, rewritten in place, so a repeatedly-drifting reviewer leaves one comment rather than a thread per run. **Read-only on the reviewer's objects**: it never edits or deletes a report, because a malformed body is evidence of what the run did and rewriting it would hide the failure this exists to surface. A violation does not fail the check by default — a malformed body doesn't mean the findings are wrong; `fail-the-check: true` opts into gating.

Bodies reach the script **through the environment, never interpolated into a `run:` block** — a review body is attacker-controlled text, and `${{ }}` inside `run:` would splice it into shell. L1 asserts this.

**`plugins/pr-reviewer-shape-guard/`** — caller template + README, following the `pr-relevance-memory` distribution pattern. No secrets; one file to copy.

**This repo is its own first consumer**, via a local `uses:`. Note that this does *not* exercise a PR’s copy: `pull_request_review` and `issue_comment` workflows — and every local `uses: ./…` they resolve — always run from the default branch, so a change to the reusable workflow takes effect here only after merge. Pre-merge coverage of the validator comes from L1 `G26`.

## Two things the classifier had to learn from real output

Both are why the fixtures live in `scripts/eval/fixtures/posted-bodies/` as **real published bodies**, not only ones I generated:

1. The flat body **omitted `**Run mode**`**, so my first classifier — requiring two diagnostic labels — missed it entirely and returned `not-a-reviewer-body`. The gate-table header is decisive on its own.
2. Column spacing varies between runs (`|---|` vs `| --- |`), so the header is matched by cells, not as a fixed string.

## Verification

`node scripts/eval/l1.mjs` → **537/537** at head (523/523 when this PR was opened; review fixes added fourteen checks).

**G26** executes the validator against the real posted bodies, asserts all three rendered snapshots **pass** it (if the renderer's own output failed the external validator, the two disagree about the contract and one is wrong), checks the ignore cases, and covers every defect class.

Mutation-tested red on every check. **One mutation initially stayed green:** the abridged flat fixture carries two other diagnostic labels, so it never exercised the gate-table-only path that I'd added specifically for it — fixed by adding the minimal synthesised case that does. Worth noting because it's the same trap as before: a fixture that's easier to detect than the real thing.

## Still open (needs a push credential with the `workflows` permission)

Six reviewer findings are validated but cannot land here — the applying identity’s token is refused on any commit touching `.github/workflows/**`:

- The self-caller header still repeats the corrected `uses:` claim (fixed in `CLAUDE.md` and above).
- `EVENT_NAME` is resolved and never passed to the validator, so a report posted to a review body still verdicts `conforms` — **the “makes the rate visible” claim below is therefore not yet met**.
- `ref: main` still overrides whatever ref a caller pins.
- The sticky-notice lookup uses `gh api --paginate` with `--jq`, which emits one object per page, so past 100 comments `JSON.parse` throws and a second notice is posted every run.
- An exit-2 validator crash leaves `verdict.json` empty, so the reporting step throws regardless of `fail-the-check: false`, and exit 2 is reported as a shape violation.
- The self-caller omits the `permissions:` block the distributed template now declares as required.

## What this does not fix

All four runs posted to **review bodies**, not the sticky — 0/4. That's the host question #121 deferred, and it's now the larger correctness problem, since duplicate reports are back by default. This guard will at least make the rate visible instead of us sampling PRs by hand.

---

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -28,6 +28,18 @@
         "url": "https://github.com/mthines/agent-skills",
         "path": "plugins/pr-relevance-memory"
       }
+    },
+    {
+      "name": "pr-reviewer-shape-guard",
+      "description": "GitHub Actions reusable workflow that validates a POSTED pr-reviewer report body against the report shape contract, from outside the agent — the only guard that sees what actually reached a PR. Install the caller template (templates/report-shape-caller.yml); no secrets required.",
+      "version": "1.0.0",
+      "author": { "name": "mthines" },
+      "license": "MIT",
+      "source": {
+        "source": "git-subdir",
+        "url": "https://github.com/mthines/agent-skills",
+        "path": "plugins/pr-reviewer-shape-guard"
+      }
     }
   ]
 }
--- a/.github/workflows/reviewer-report-shape-self.yml
+++ b/.github/workflows/reviewer-report-shape-self.yml
@@ -1,0 +1,45 @@
+# pr-reviewer report shape guard — this repo's own caller.
+#
+# Generated from plugins/pr-reviewer-shape-guard/templates/report-shape-caller.yml, with the
+# `uses:` pointed at the local path instead of @main so a change to the reusable workflow is
+# exercised by the PR that makes it, rather than only after merge.
+#
+# Copy this file to:
+#   YOUR_REPO/.github/workflows/reviewer-report-shape.yml
+#
+# No secrets required. That is the entire setup.
+#
+# What this does:
+#   Every time pr-reviewer posts a review or a comment, this checks the POSTED body against the
+#   report shape contract — marker present, `Review details` accordion present and collapsed,
+#   nothing the accordion owns rendered above it, no advisory verdict, no link caged in a code
+#   span, and every declared count equal to the list it counts.
+#
+#   On a violation it leaves ONE sticky notice on the PR naming exactly what is malformed. It never
+#   edits or deletes the report itself.
+#
+# Why you want it: every other guard on this contract runs inside the agent's own control flow, so a
+# run that hand-writes the body is unobserved. This is the only check that sees what was published.
+#
+# See: https://github.com/mthines/agent-skills/blob/main/plugins/pr-reviewer-shape-guard/README.md
+
+name: Reviewer report shape
+
+on:
+  pull_request_review:
+    types: [submitted, edited]
+  issue_comment:
+    types: [created, edited]
+
+jobs:
+  shape:
+    # issue_comment fires on plain issues too; PR-only, and skip humans' own comments cheaply.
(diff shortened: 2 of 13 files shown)


mthines:
@dash0 implement

PR #136, opened 2026-08-25
feat(severity): severity skill + severity-aware reviewer thresholds (default)

mthines:
## What

Adds a `severity` skill — a standalone quality primitive that rates how bad a finding or bug is **if it's real** (`critical`/`high`/`medium`/`low`), the blast-radius axis complementary to `confidence`'s is-it-real axis. Then wires it into the reviewer as an **opt-in, back-compatible** severity-aware confidence threshold.

## Why

Dogfooding our PR-reviewer against Cursor Bugbot across ~600 dash0hq/dash0 PRs showed the gap isn't our confidence bar — it's that we don't rate or route by severity. Cursor's tiers are meaningful (independent corroboration fell 41% → 33% → 19% down High → Medium → Low). Lowering the single confidence knob would just add false positives; the right lever is severity, so severe findings surface at a lower confidence bar and low-value nits stay advisory.

## Increasing findings — two levers (be precise)

- **This PR (routing):** severity-by-default *promotes* moderately-confident high/critical findings from the advisory band into inline, and *demotes* low nits. More signal, less noise — better **composition**.
- **Follow-up (recall):** matching Cursor's ~2x raw volume needs the deep bug-hunting pass so the reviewer *generates* more real bugs; a threshold can only surface what was found. Tracked separately.

## How it's shaped (mirrors `confidence`)

- **Emits a tier, nothing else.** Policy-free — the numeric bars live in the consumer, not the skill.
- **Honest determinism.** Exactly one thing is deterministic: an executable path floor (`auth`/`billing`/`migration`/`infra`/`secrets` globs). The data-loss/security/concurrency escalators are labelled heuristic (LLM-judged), not deterministic.
- **No noise inflation.** An exclusion gate runs first — floors never fire on test/fixture/generated/reverse-migration paths — and a reachability cap rates down unreachable code. These directly answer the two hardest findings from the skill's own `/critical` review.
- **`(blocking)` crosswalk.** `critical`/`high` ⇒ `(blocking)`; feeds the reviewer's existing decoration rather than forking a second severity system.

## Wiring (opt-in, back-compat)

- `review-config.md`: `severity_thresholds` map, **on by default**. `medium` anchors each profile's historical bar (balanced = 80, unchanged for a typical finding); `critical`/`high` get a lower bar (surface more), `low` a higher one (surface less). A flat `per_comment_confidence_threshold` override restores the uniform bar.
- `per-comment-confidence.md`: the reviewer tiers every finding via `Skill("severity","finding")` in the same pass (no extra per-finding call) and gates it on the per-tier bar.

## Tests

- L1 deterministic contract checks: **685/685** pass (new `severity-tiering` suite registered, anchors/links clean).
- L2 `severity-tiering` suite added (15 golden cases incl. the exclusion and reachability tricky-cases). Runs in CI via `evals-l2.yml` on this PR's golden/rubric changes.

## Severity is visible

- **Inline:** each finding shows its tier as a Conventional-Comments label, e.g. `issue (high): …`. The load-bearing `(blocking)`/`(non-blocking)` token is untouched.
- **Report:** a `**Severity**` line renders a glyph breakdown — `🔴 1 · 🟠 2 · 🟡 3 · ⚪ 1` — via the fail-closed renderer (optional `TIER_TALLY` payload; existing fixtures unchanged, `pass` snapshot regenerated).

## Severity is recorded end-to-end

The reviewer leaves a hidden `<!-- severity: <tier> -->` marker on each tiered finding (excluded from comment-shape caps); `record-comment-relevance.mjs` strips it before fingerprinting (verified fingerprint-stable) and reads it into the `reviewer-comment-relevance` record's optional `severity` field. Per-tier precision is now measurable over time.

## Deliberately deferred

Provenance: severity skill authored via `/create-skill`, reviewed via `/critical` (all findings fixed), gated via `/confidence`.


agents/shared/rules/conventional-comments.md
```diff
@@ -69,13 +80,17 @@ question: Is the empty `catch {}` intentional? Curious whether we want to surfac
 Before any emit:
 
 ```python
+import re
+
+# Tolerates the optional severity label decoration, e.g. "issue (high):".
+PREFIX_RE = re.compile(
```
mthines, line 86: issue (high): `PREFIX_RE` now tolerates the `(tier)` label, but the Step 4b `payload_is_safe` pre-flight still requires `startswith("issue:")`, so it rejects `issue (high):` and aborts on the reviewer's own tiered comments. **(blocking)**

agents/shared/rules/per-comment-confidence.md
```diff
@@ -34,6 +34,18 @@ For each finding that survives `finding-grounding.md` (2.6) and `verification-re
    The default (no review config present or `profile: balanced`) is **80**.
    See `review-config.md` for the full profile→threshold table.
 
+   **Severity-aware threshold (default).** Assess the finding's tier once with
+   `Skill("severity", "finding")` — in this same pass as the confidence rating, never a
+   separate per-finding call — and use the profile's `severity_thresholds[tier]` as the
```
mthines, line 39: issue (medium): This sets the effective bar per severity tier, but the per-type table (`issue` 70, `suggestion` 90, `nitpick` 95) still stands — a `suggestion` tiered `high` has two conflicting bars (90 vs 70). **(non-blocking)**

skills/quality/severity/SKILL.md
```diff
@@ -0,0 +1,265 @@
...
+(`agents/shared/rules/conventional-comments.md`), a cap-exemption that always posts a
+`(blocking)` finding inline (`rubric-composition.md § Placement`), and an orthogonal
+`materiality` dimension. This skill does not replace them — it feeds the `(blocking)`
+decision. The crosswalk (note: a tier is orthogonal to the `issue`/`suggestion`/`nitpick`
+category, which `rubric-composition.md § Severity mapping` still assigns separately):
+
+| Tier         | Reviewer treatment                                                                 |
+| ------------ | --------------------------------------------------------------------------------- |
+| **critical** | `(blocking)` — cap-exempt, always inline.                                          |
+| **high**     | `(blocking)` — cap-exempt, always inline.                                          |
```
mthines, line 221: issue (medium): The `high` ⇒ `(blocking)` crosswalk fires even when `high` came only from the Step 2 path floor, so a cosmetic finding on an `auth`/`billing`/`migration` path becomes blocking and flips Gate 6 to ❌. **(non-blocking)**

agents/shared/rules/per-comment-confidence.md
```diff
@@ -34,6 +34,18 @@ For each finding that survives `finding-grounding.md` (2.6) and `verification-re
...
+   effective threshold (`review-config.md` § Severity-aware thresholds). The `medium` tier
+   anchors the profile's historical bar (**80** for `balanced`), so a typical finding is
+   gated exactly as before; `critical` / `high` get a lower bar (surface more) and `low` a
+   higher one (surface less). A flat `per_comment_confidence_threshold: N` override
+   collapses all tiers back to `N` (pre-severity behavior). The `defer_floor` band and all
+   rules below use the effective threshold unchanged. A finding tiered `critical` or `high`
+   is additionally decorated `(blocking)` (severity's § Mapping to a reviewer's blocking
```
mthines: issue (medium): The floor-only `(blocking)` exception (`severity/SKILL.md`) is not mirrored here — this still marks every `critical`/`high` `(blocking)` + cap-exempt, undoing fix #4 (same at `review-config.md:128`). **(non-blocking)**

PR #137, opened 2026-08-27
feat(pr-reviewer): opt-in Fix-with-Agent0 deep-link buttons

mthines:
## What

Opt-in **"Fix with Agent0"** deep-link buttons on the reviewer's output — the Agent0 equivalent of Cursor's "Fix in Cursor" button. Two placements: a **Fix all** button in the report and a **Fix this** button per inline finding. Off by default.

> **Stacked on #136** (`feat/severity-skill`). It extends the same files the severity work touches (`render-report.mjs`, `comment-shape.md`, `review-config.md`), so `main` isn't the right base yet. GitHub retargets this to `main` when #136 merges; review/merge #136 first.

## How

- **Deep link:** `https://app.dash0.com/goto/agent0?auto_submit=true&initial_prompt=<encoded>`. A small deterministic encoder (`scripts/build-agent0-link.mjs`) does `encodeURIComponent` **plus** `( ) ' → %28 %29 %27`, because those stay literal otherwise and a bare `)` would terminate the `](url)` markdown link. Single source, so the report and inline sites encode identically.
- **Prompts (grounded in `/ai-engineering`):** *Fix this* embeds only the reviewer's own finding, wrapped in `<finding>` delimiters and labelled as task-data. *Fix all* embeds **no** comment text — it points Agent0 at the PR and is scoped to *the reviewer's own report + its own threads*, explicitly ignoring other authors/bots. That keeps untrusted comment text out of an auto-submit URL.
- **Buttons:** two committed SVGs (`agents/pr-reviewer/assets/`). Referenced by a configurable URL (default = the repo raw asset); for production Dash0 should host PNGs on `app.dash0.com` like Cursor does — noted in the rule.
- **Opt-in is a runtime mode:** the `--fix-links` flag on the `pr-reviewer` invocation (matching `--critical`/`--full`/`--no-*`), which the Agent0 automation passes per-run — nobody else's reviews change and no repo commits a config file. `agent0_fix_links: true` in `.github/review.yaml` stays as an optional repo-wide default. Default off.

## Rendering

- **Report:** optional `FIX_ALL_URL` payload slot → the renderer builds the linked button above the accordion, fail-closed (validated `http(s)`, rejects a literal `)`), rendered only when supplied. Existing `warn`/`fail` fixtures are byte-identical (field optional); `pass` snapshot regenerated.
- **Inline:** the agent appends the *Fix this* button after the fix block on `issue:`/`suggestion:` findings when the flag is on (`comment-shape.md`), skipped for `nitpick`/`question`/`praise`.

## Tests

- L1 **688/688**; renderer `node --check` clean; all three report snapshots match; encoder self-tested.

## Safety

The buttons only *prepare* a prompt — a human clicks, and Agent0 runs under its own guardrails and commits to the PR the human is viewing. The reviewer never triggers a fix itself. *Fix all* names the reviewer's own report by marker and ignores every other author, so a third-party comment can't ride into the run.


mthines:
@dash0 implement

mthines:
@dash0 exec Can you create a new pr which adds the .github/review.yml and agent0_fix_links:true config to this repo and the lorekit repo?

PR #165, opened 2026-09-03
One comment system across the pr-reviewer report and inline findings

mthines:
The report and the inline findings looked like two different bots, because only one of them had a renderer.

`render-report.mjs` exists because runs stopped copying the report template and started remembering it. `comment-shape.md` had the identical problem and the opposite treatment — prose rules plus a `passes_shape()` Python function that nothing executed — so the report's contract held while the inline one eroded. On `dash0hq/dash0#18362` a posted finding dropped three documented decorations at once (the severity label, the bold on `**(blocking)**`, and its position) and omitted the required fix fence, and the report headline (`Not ready — this PR adds github.check_run…`) matched none of the three forms `report-rendering.md § Headlines` defines — `HEADLINE` was the most-read line in the system and the only slot checked merely for non-emptiness.

## What changed

**One shared spine.** New `agents/pr-reviewer/scripts/comment-spine.mjs` owns the tier glyph map, the attribution footer builder, the theme-aware button markup, and the caps. Both renderers import it, and that single import is the mechanism: a change to the vocabulary cannot land on one surface only.

**An inline renderer.** New `agents/pr-reviewer/scripts/render-comment.mjs` is the counterpart to `render-report.mjs` — one JSON payload per finding in, one comment body out, fail-closed on an unknown key or a violated cap, with committed reference renderings in `scripts/eval/fixtures/inline-comment/`. A claim now opens `issue (high): 🟠 **<≤ 60-char noun phrase>** **(blocking)**`; the prefix, tier label and colon keep position 0 so `record-comment-relevance.mjs`'s `SEVERITY_RE` reads the tier unchanged. The title exists because GitHub shows only a comment's opening in the Files rail, the timeline and the notification email. A one-liner (`nitpick:` / `question:` / `praise:`) stays one line, and the renderer rejects a title on one and requires it on the other. Prose drops to 200 chars to pay for the title.

**One asymmetry, stated as a rule:** a `###` heading means the report, a bold title means one finding. Each renderer's post-conditions enforce its own half.

**`HEADLINE` is gone.** `VERDICT` + `FINDINGS[]` + `SUMMARY` replace it, and the renderer derives the count-forward heading (`### 🟠 4 findings — 1 blocking`), a visible findings index above the accordions (one linked row per inline finding, carrying the same title — the worklist the report never had), and the `Severity — 🟠 1 high` tally with the glyph paired with its word (WCAG 1.4.1). Contradictions the report could previously post became unrepresentable: a `VERDICT` disagreeing with its own gate table, a `FINDINGS` length disagreeing with `QUALITY`'s `posted inline <N>`, and — since `9ae09c3` — a reasons line claiming more warnings than the table has ⚠️ gates.

**The footer is shared verbatim and moved out of the accordion.** `<sup>`pr-reviewer` · commit `<sha7>` · …</sup>` is the cue that makes an inline finding and the report recognisably the same reviewer, and it is the only attribution visible in a notification email. Before this, the report's was collapsed inside `Review details` and the inline comments had none.

**Reader-facing labels lost the pipeline vocabulary.** `cleared review, not inlined` → `<N> more findings — verified, too minor to comment on`; `Low-confidence findings` → `Less certain`.

**Fix buttons are on by default** wherever the review config names an `agent0_environment` (`--no-fix-links` opts out; one flag governs both placements, so a run has both buttons or neither). Rendered from new theme-aware 36 px assets via `<picture>` — up from a 28 px dark-only chip that was the primary CTA of the review and ignored the reader's theme. The 28 px assets are kept and unreferenced so already-posted comments do not 404.

## Two failure modes found by posting, not by reasoning

Both were discovered by running the reviewer against this PR and reading the stored bodies back. Neither was predictable from the renderer, which is the point: every guard before this ran inside the agent's control flow.

**A relay mangles long URLs, and no markup evades it.** A write path that carries the body as a tool-call argument wraps any URL over ~140 body chars in a code span, closing the `href` and escaping the markup after it. Nine probes established that length is the whole trigger — host, path, `auto_submit=true`, and prompt-like query content are each innocent, and a pure-markdown linked image is rewritten too. This **falsified** the first diagnosis, which had blamed the model reformatting a long HTML line while copying it. The remedy is `--relay-check` and withholding the buttons, never shortening the prompt: a `fix-this` link spends 106 body chars before the prompt starts, and a button that opens a session with no idea what to fix is worse than none. `--relay-check` exits 1 only on a *fix link* over budget and 3 on any other long URL, because `--no-fix-links` removes Agent0 deep links and nothing else — reporting a cited doc URL as exit 1 was a gate with no exit.

**A `<picture>` does not fall back to ``.** `ASSET_BASE` is pinned to the default branch, so the four theme variants this PR adds 404 until it merges, and both `<source>` elements failing renders a broken-image icon rather than the default. `--assets-check` HEADs each asset and exits 1 (withhold), 0 (post), or 3 (network unreachable — inconclusive, post as rendered; an unreachable network must never withhold a button that would have rendered). The unsuffixed `{stem}.svg` is now a byte-copy of the light variant, because notification email and RSS ignore `<source>` entirely and render the default.

**Consequence, stated plainly: the buttons do not render on this PR.** Both guards fire correctly on every run against it, and they will render on the first review after this merges. Pointing `ASSET_BASE` at a feature branch to make a screenshot look right is documented as not a fix.

## Two deliberate calls worth a reviewer's attention

- **The button anchor is HTML `<a href>`, not markdown link brackets around `<picture>`.** Whether a markdown parser treats a block of raw HTML as a link's inline content is not something this repo can verify without posting a comment, and a broken button reads as a broken reviewer. Both elements are on GitHub's comment HTML allowlist.
- **The terminal/posted-headline byte-identity rule is retired.** The posted headline is now derived and the terminal one is not, so byte identity was no longer expressible. The surviving contract is that the verdict token and reason phrases agree, enforced by the gate-table cross-check.

## Verification

- `node scripts/eval/l1.mjs` — 1063 → **1224 checks, all green**. New `G46` block (a–m): scripts exist, self-test passes, fixtures round-trip, per-snapshot invariants (prefix at position 0, the recorder's verbatim `SEVERITY_RE`, bold decoration on line 1, shared footer, no heading, title present iff claim), both renderers import the spine and hold no local copies (`G46f`), the tally pairs glyph with word, fail-closed rejections, and `G46i` extracts `payload_is_safe` from the agent body and executes it against the reference renderings.
- `node agents/pr-reviewer/scripts/render-comment.mjs --self-test` — **39/39**.
- **Every guard in this change was probed by breaking the thing it protects** — 20 probes across `G46g`–`G46m`, each confirmed to fail correctly and then restored. Two of them caught real divergences during development: `G46i` found the `<picture>` markup being measured as prose (so `payload_is_safe` rejected every rendered claim), and `G46j` found a guard that counted `--relay-check` occurrences and therefore passed with a call site deleted, because the body mentions the flag in prose too. Call sites are now anchored by their target file.
- `G46l` measures a maximal legal payload for each shape against the ceiling it reads out of `payload_is_safe` itself, so it catches the class rather than the instance: raising `PROSE_MAX` fails it. It exists because `UNVERIFIED_MAX` was added to the spine without that ceiling moving, and a finding legal under every per-field cap measured 363 against a 320 bound — a predicate that aborts the *whole* post, so one maximal finding took an entire review down.
- `scripts/validate-report-shape.mjs` gained `no-headline-heading` and `findings-index-inside-accordion`, both scoped to bodies carrying the new footer so already-posted reports are unaffected.
- Also on this branch: `59d96d2` closes the `pr-reviewer` memory record schemas against invented fields (cherry-picked from `claude/pr-reviewer-quality-1kn8ba` so closing that branch's PR did not park it — independent of everything else here, and revertable as one commit).

https://claude.ai/code/session_015z8UFRtJGJXoUd66beoNRz

---
_Generated by [Claude Code](https://claude.ai/code/session_015z8UFRtJGJXoUd66beoNRz)_


mthines:
### 🟠 5 findings — 2 blocking

Gives the inline comment surface its own fail-closed renderer and moves the glyphs, footer, caps, and button markup into one spine that both renderers import.

**Blocking:** 2 blocking findings (see inline)

|---|---|---|
| TITLE cannot name a dotted symbol the report renders fine | `agents/pr-reviewer/scripts/render-comment.mjs:124` | 🟠 high · blocking |
| Uncapped UNVERIFIED tag can abort the whole post | `agents/pr-reviewer/scripts/render-comment.mjs:165` | 🟠 high · blocking |
| Second G41 block reuses all eight sub-IDs of the first | `scripts/eval/l1.mjs:3319` | 🟡 medium |
| Fix all count reaches only the alt text, not the button | `agents/pr-reviewer/scripts/render-report.mjs:988` | ⚪ low |
| Stale default false clause contradicts the new omit rule | `agents/shared/rules/review-config.md:74` | ⚪ low |

<a href="``https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=Fix%20the%205%20open%20pr-reviewer%20findings%20on%20mthines%2Fagent-skills%23165%2C%20authored%20by%20mthines.%20List%20them%20with%3A%20gh%20api%20graphql%20-f%20query%3D%27%7Brepository%28owner%3A%22mthines%22%2Cname%3A%22agent-skills%22%29%7BpullRequest%28number%3A165%29%7BreviewThreads%28first%3A100%29%7BpageInfo%7BhasNextPage%20endCursor%7D%20nodes%7BisResolved%20comments%28first%3A1%29%7Bnodes%7Bpath%20line%20body%20author%7Blogin%7D%7D%7D%7D%7D%7D%7D%7D%27%20%E2%80%94%20if%20hasNextPage%2C%20repeat%20with%20after%3A%22%3CendCursor%3E%22%20until%20false.%20Keep%20only%20isResolved%3Afalse%20nodes%20whose%20comment%20author.login%20is%20%22mthines%22%2C%20ignore%20every%20other%20author.%20Verify%20using%20the%20repo%27s%20own%20lint%2Ftypecheck%20scripts%20%E2%80%94%20never%20a%20raw%20tsc%2Feslint%20call%20or%20a%20whole-repo%20pass%20%E2%80%94%20skip%20verification%20if%20none%20exist%20%E2%80%94%20then%20commit%20to%20the%20same%20branch%20%28no%20new%20PR%29.&amp;utm_source=pr-reviewer-fix-all"&gt;&lt;picture&gt;&lt;source`` media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-all-agent0-dark.svg">Fix all 5 with Agent0</picture></a>

<details>

**Needs attention**

| Gate | Status | Details |
|---|---|---|
| Description vs. code | ✅ | The description matches what the diff does. |
| Prior review feedback | ✅ | Earlier review comments are resolved. |
| Documentation | ✅ | The change is documented well enough to follow. |
| Self-review signals | ✅ | No debug logs, leftover TODOs, or unreviewed stubs. |
| Code review | ❌ | 2 blocking findings in the new inline renderer, plus 3 non-blocking — see the inline comments. |

**Found**

Quality — produced 7 → posted inline 5; carried forward 0, filter drops 0, materiality drops 0, dedupe drops 0, grounding drops 0, verifier contradicted 2, confidence drops 0 (threshold 80), confidence-deferred 0, shape drops 0, memory suppressions 0, cleared 5, deferred over inline cap 0
Severity — 🟠 2 high · 🟡 1 medium · ⚪ 2 low
Verified — Traced both consumers of the new `assertPlain`; re-ran l1 (1134 checks) and render-comment --self-test (37/37); both GitHub checks green; probed each renderer cross-check for a non-zero exit with empty stdout.

**Run**

full · 2498 lines in delta · tier deep · depth checkout · 41 files touched · blast_radius=high · semver_delta=none · shapes error-handling, propagation
⚠️ Reviewed at HEAD 82f8ac8, not the working tree — the four button assets carry uncommitted local edits; and blast_radius=high is graph-inflated by generic identifiers (main, build, anchor) matching repo-wide, so it is treated as a lead.
Memories — 10 indexed · 2 used (1 knowledge · 1 hotspot)

- **hotspot** [`hotspot::agents/pr-reviewer.md`](``https://lorekit.io/lore?scope=%22repo%3A%3Amthines%2Fagent-skills%22&lesson=%7B%22scope%22%3A%22repo%3A%3Amthines%2Fagent-skills%22%2C%22key%22%3A%22hotspot%3A%3Aagents%2Fpr-reviewer.md%22%7D)`` — 3 prior confirmed here — a rule file gains a section while the body's step, budget, or slot name does not; targeted this run's seam check
- **knowledge** ``['knowledge::INTEGRATIONS@agents/pr-reviewer/scripts/render-report.mjs'](https://lorekit.io/lore?scope=%22repo%3A%3Amthines%2Fagent-skills%22&lesson=%7B%22scope%22%3A%22repo%3A%3Amthines%2Fagent-skills%22%2C%22key%22%3A%22knowledge%3A%3AINTEGRATIONS%40agents%2Fpr-reviewer%2Fscripts%2Frender-report.mjs%22%7D)`` — the renderer fails closed with zero bytes on stdout, and used may never exceed indexed

<sup>Nothing to report — standards (1 doc), optimality (3 judged), measurability (skipped), integrations (not activated), 0 files skipped.</sup>

</details>

<sup>`pr-reviewer` · commit `82f8ac8` · full review · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md) · updated 2026-09-03 19:39 UTC</sup>


---

agents/pr-reviewer/scripts/render-comment.mjs
```diff
@@ -0,0 +1,485 @@
...
+  // none of them a handle. A `nitpick` is the opposite case: a title would cost more height than
+  // the nitpick is worth, which is why the terse form survives unchanged.
+  let title = null;
+  if (isClaim) {
+    if (!data.TITLE || String(data.TITLE).trim() === "") {
+      bad(`TITLE is required on a ${prefix}: — it is what a reader sees in the Files rail and in a`
+        + " notification email, where the body is not visible");
+    }
+    title = String(data.TITLE).trim();
+    assertPlain("TITLE", title);
```
mthines: issue (high): 🟠 **TITLE cannot name a dotted symbol the report renders fine** **(blocking)**

Backticks are rejected by `assertPlain` and the bare form then trips `sentenceCount`, so both routes are closed — while `render-report.mjs` passes `allowCode: true` on the same string.

Evidence: `agents/pr-reviewer/scripts/render-report.mjs:570` (allowCode true) · `scripts/eval/fixtures/report-body/deep.json:9` (both routes rejected)

```js
assertPlain("TITLE", title, { allowCode: true });   // parity with render-report.mjs:570
```

<a href="``https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=Fix%20the%20pr-reviewer%20finding%20at%20agents%2Fpr-reviewer%2Fscripts%2Frender-comment.mjs%3A124%20on%20mthines%2Fagent-skills%23165%20%E2%80%94%20%22issue%20%28high%29%3A%20TITLE%20cannot%20name%20a%20dotted%20symbol%20the%20report%20renders%20fine%22.%20Read%20the%20inline%20comment%20there%20for%20detail.%20Verify%20using%20the%20repo%27s%20own%20lint%2Ftypecheck%20scripts%20%E2%80%94%20never%20a%20raw%20tsc%2Feslint%20call%20or%20a%20whole-repo%20pass%20%E2%80%94%20skip%20verification%20if%20none%20exist%20%E2%80%94%20then%20commit%20to%20the%20same%20branch%20%28no%20new%20PR%29.&amp;utm_source=pr-reviewer-fix-this"&gt;&lt;picture&gt;&lt;source`` media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg">Fix with Agent0</picture></a>

<sup>`pr-reviewer` · commit `82f8ac8` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

agents/pr-reviewer/scripts/render-comment.mjs
```diff
@@ -0,0 +1,485 @@
...
+  //
+  // `EVIDENCE` and `UNVERIFIED` are mutually exclusive by construction, because they are opposite
+  // claims: one cites what a receipt touched, the other says no receipt was obtainable. A comment
+  // carrying both asserts a proof it also says it does not have
+  // (`verification-receipt.md § unobtainable`).
+  let unverified = null;
+  if (data.UNVERIFIED !== undefined && data.UNVERIFIED !== null
+      && String(data.UNVERIFIED).trim() !== "") {
+    unverified = String(data.UNVERIFIED).trim();
+    assertPlain("UNVERIFIED", unverified, { allowCode: true });
```
mthines, line 181: issue (high): 🟠 **Uncapped UNVERIFIED tag can abort the whole post** **(blocking)**

It renders on line 1 outside every prose cap, so a 107-char reason clears the renderer and then fails `payload_is_safe` at 355 chars — which aborts the whole review rather than dropping one comment.

Evidence: `agents/pr-reviewer.md:2785` (the 320-char ceiling)

```js
if (unverified.length > 80) {
  bad(`UNVERIFIED is ${unverified.length} chars — it rides on line 1, inside the 320-char pre-flight budget`);
}
```

<a href="``https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=Fix%20the%20pr-reviewer%20finding%20at%20agents%2Fpr-reviewer%2Fscripts%2Frender-comment.mjs%3A165%20on%20mthines%2Fagent-skills%23165%20%E2%80%94%20%22issue%20%28high%29%3A%20Uncapped%20UNVERIFIED%20tag%20can%20abort%20the%20whole%20post%22.%20Read%20the%20inline%20comment%20there%20for%20detail.%20Verify%20using%20the%20repo%27s%20own%20lint%2Ftypecheck%20scripts%20%E2%80%94%20never%20a%20raw%20tsc%2Feslint%20call%20or%20a%20whole-repo%20pass%20%E2%80%94%20skip%20verification%20if%20none%20exist%20%E2%80%94%20then%20commit%20to%20the%20same%20branch%20%28no%20new%20PR%29.&amp;utm_source=pr-reviewer-fix-this"&gt;&lt;picture&gt;&lt;source`` media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg">Fix with Agent0</picture></a>

<sup>`pr-reviewer` · commit `82f8ac8` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

scripts/eval/l1.mjs
```diff
@@ -3198,6 +3316,167 @@ const isPollBlock = (block) =>
   }
 }
 
+// ── G41: the inline comment surface has a renderer, and it shares ONE vocabulary with the report ──
```
mthines: suggestion (medium): 🟡 **Second G41 block reuses all eight sub-IDs of the first** **(non-blocking)**

G41a through G41h are each claimed by both blocks, so a failing check ID no longer locates its check and the `G41f` citation in `CLAUDE.md` resolves to the wrong one.

Evidence: `scripts/eval/l1.mjs:3566` (the first G41 block) · `CLAUDE.md:110` (cites G41f)

```js
// ── G45: the inline comment surface has a renderer, and it shares ONE vocabulary … ──
//   …and rename this block's G41a–G41i checks to G45a–G45i.
```

<a href="``https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=Fix%20the%20pr-reviewer%20finding%20at%20scripts%2Feval%2Fl1.mjs%3A3319%20on%20mthines%2Fagent-skills%23165%20%E2%80%94%20%22suggestion%20%28medium%29%3A%20Second%20G41%20block%20reuses%20all%20eight%20sub-IDs%20of%20the%20first%22.%20Read%20the%20inline%20comment%20there%20for%20detail.%20Verify%20using%20the%20repo%27s%20own%20lint%2Ftypecheck%20scripts%20%E2%80%94%20never%20a%20raw%20tsc%2Feslint%20call%20or%20a%20whole-repo%20pass%20%E2%80%94%20skip%20verification%20if%20none%20exist%20%E2%80%94%20then%20commit%20to%20the%20same%20branch%20%28no%20new%20PR%29.&amp;utm_source=pr-reviewer-fix-this"&gt;&lt;picture&gt;&lt;source`` media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg">Fix with Agent0</picture></a>

<sup>`pr-reviewer` · commit `82f8ac8` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

agents/pr-reviewer/scripts/render-report.mjs
```diff
@@ -835,24 +889,120 @@ function main() {
...
-    fixAllButton = `[![Fix all with Agent0](${ASSET})](${u})`;
+    // Zero findings and no CI note means there is nothing for Agent0 to do, and a button that
+    // hands it an empty worklist is worse than no button: it invites a click that spends a run
+    // discovering it has no work. The one legitimate zero-finding case is the CI-only template
+    // (`agent0-fix-links.md § Fix all — CI-only`), which requires a CI note to exist.
+    if (n === 0 && !data.CI_NOTE) {
+      fail("FIX_ALL_URL with 0 findings and no CI_NOTE — there is nothing to fix; omit the button"
+        + " (agent0-fix-links.md § Fix all — CI-only)");
+    }
+    const label = n > 0 ? `Fix all ${n} with Agent0`
```
mthines: suggestion (low): ⚪ **Fix all count reaches only the alt text, not the button** **(non-blocking)**

The count lands in `alt` while the asset bakes in `Fix all with Agent0`, so the promise the comment above says a reader can check never reaches a sighted one.

Evidence: `agents/pr-reviewer/assets/fix-all-agent0-light.svg:10` (the baked label) · `agents/shared/rules/agent0-fix-links.md:326` (documents the count)

```js
// The label carries the count in `alt` only — the asset's own text is fixed, so the
// count is announced to a screen reader but is not visible on the button itself.
```

<a href="``https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=Fix%20the%20pr-reviewer%20finding%20at%20agents%2Fpr-reviewer%2Fscripts%2Frender-report.mjs%3A988%20on%20mthines%2Fagent-skills%23165%20%E2%80%94%20%22suggestion%20%28low%29%3A%20Fix%20all%20count%20reaches%20only%20the%20alt%20text%2C%20not%20the%20button%22.%20Read%20the%20inline%20comment%20there%20for%20detail.%20Verify%20using%20the%20repo%27s%20own%20lint%2Ftypecheck%20scripts%20%E2%80%94%20never%20a%20raw%20tsc%2Feslint%20call%20or%20a%20whole-repo%20pass%20%E2%80%94%20skip%20verification%20if%20none%20exist%20%E2%80%94%20then%20commit%20to%20the%20same%20branch%20%28no%20new%20PR%29.&amp;utm_source=pr-reviewer-fix-this"&gt;&lt;picture&gt;&lt;source`` media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg">Fix with Agent0</picture></a>

<sup>`pr-reviewer` · commit `82f8ac8` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

agents/shared/rules/review-config.md
```diff
@@ -67,8 +67,11 @@ severity_thresholds:                     # DEFAULT — values shown are the `bal
                                          # A flat per_comment_confidence_threshold: N
                                          # override collapses all tiers back to N.
 
-agent0_fix_links: false                  # repo-wide default for the "Fix with Agent0" buttons —
-                                         # equivalent to always passing the --fix-links mode
+agent0_fix_links: false                  # explicit override for the "Fix with Agent0" buttons —
+                                         # omit it and the buttons follow agent0_environment
+                                         # (named ⇒ on, absent ⇒ off) —
+                                         # true is equivalent to always passing --fix-links,
+                                         # false to always passing --no-fix-links
```
mthines, line 74: suggestion (low): ⚪ **Stale default false clause contradicts the new omit rule** **(non-blocking)**

The added lines say an omitted key follows `agent0_environment`, and the retained clause still says the default is false, so one comment block states both rules.

Evidence: `agents/shared/rules/review-config.md:75` (the retained clause)

```yaml
                                         # false to always passing --no-fix-links;
                                         # see agents/shared/rules/agent0-fix-links.md
```

<a href="``https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=Fix%20the%20pr-reviewer%20finding%20at%20agents%2Fshared%2Frules%2Freview-config.md%3A74%20on%20mthines%2Fagent-skills%23165%20%E2%80%94%20%22suggestion%20%28low%29%3A%20Stale%20default%20false%20clause%20contradicts%20the%20new%20omit%20rule%22.%20Read%20the%20inline%20comment%20there%20for%20detail.%20Verify%20using%20the%20repo%27s%20own%20lint%2Ftypecheck%20scripts%20%E2%80%94%20never%20a%20raw%20tsc%2Feslint%20call%20or%20a%20whole-repo%20pass%20%E2%80%94%20skip%20verification%20if%20none%20exist%20%E2%80%94%20then%20commit%20to%20the%20same%20branch%20%28no%20new%20PR%29.&amp;utm_source=pr-reviewer-fix-this"&gt;&lt;picture&gt;&lt;source`` media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg">Fix with Agent0</picture></a>

<sup>`pr-reviewer` · commit `82f8ac8` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

mthines (commented):
---

mthines:
**Write-path verification** — not a review finding. This comment carries a real `render-comment.mjs` body, reproduced byte-for-byte from the renderer's output, to check that the MCP write path no longer mangles it (the defect on this PR's earlier report and its five inline comments). The finding it describes is already fixed in `2acfa9f`.

---

issue (high): 🟠 **Backtick in an href kills the button** **(blocking)**

A backtick past position 0 was accepted by `assertBareUrl`, and inside an `href` it opens a code span that escapes the surrounding markup.

Evidence: `agents/pr-reviewer/scripts/comment-spine.mjs:180` (whitelist)

```js
const illegal = [...u].find((c) => !/[A-Za-z0-9\-._~:\/?#[\]@!$&'*+,;=%]/.test(c));
```

<a href="``https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=Fix%20the%20finding%20at%20agents%2Fpr-reviewer%2Fscripts%2Fcomment-spine.mjs%3A180%20on%20mthines%2Fagent-skills%23165.&amp;utm_source=pr-reviewer-fix-this"&gt;&lt;picture&gt;&lt;source`` media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg">Fix with Agent0</picture></a>

<sup>`pr-reviewer` · commit `2acfa9f` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

mthines:
**Write-path probe** (2 of 2) — three ways to render the same button, to find which survives this write path. Disposable; delete both probe comments whenever.

A. raw anchor + picture (what the renderer emits today):

<a href="https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;utm_source=probe-a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg">Fix with Agent0</picture></a>

B. raw anchor + plain img (no picture element):

<a href="https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;utm_source=probe-b">Fix with Agent0</a>

C. pure markdown image link (no HTML at all):

[Fix with Agent0](https://app.dash0-dev.com/goto/agent0?auto_submit=true&utm_source=probe-c)`

---

mthines:
**Write-path probe** (3 of 3) — separating URL *length* from URL *content*. Disposable.

D. long URL, innocuous padding only (no instruction-like text):

<a href="`https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;padding=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&amp;utm_source=probe-d"&gt;&lt;img` alt="probe D" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-light.svg" height="36"></a>

E. short URL, but carrying instruction-like prompt text:

<a href="`https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=Fix%20the%20finding%20and%20commit%20to%20the%20same%20branch&amp;utm_source=probe-e"&gt;&lt;img` alt="probe E" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-light.svg" height="36"></a>

---

mthines:
**Write-path probe (iteration 1)** — one variable per row. Markup is held constant at variant B's shape (raw `<a href>` + plain ``), which already survived. Disposable.

F. Agent0 host+path, **no** `auto_submit`, **no** `initial_prompt`:

<a href="https://app.dash0-dev.com/goto/agent0?utm_source=probe-f">probe F</a>

G. innocuous host, **with** `auto_submit` + `initial_prompt`:

<a href="https://example.com/goto/agent0?auto_submit=true&amp;initial_prompt=Fix%20the%20finding&amp;utm_source=probe-g">probe G</a>

H. Agent0 host+path, `auto_submit=true` **only**:

<a href="https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;utm_source=probe-h">probe H</a>

I. Agent0 host+path, `initial_prompt` **only**, three characters:

<a href="https://app.dash0-dev.com/goto/agent0?initial_prompt=Fix&amp;utm_source=probe-i">probe I</a>

---

mthines:
**Write-path probe (iteration 2)** — bisecting the URL-length threshold. Every URL below is on an innocuous host with innocuous `padding=` content, so **length is the only variable**. Disposable.

**110 chars** (URL length `110`):

<a href="https://example.com/a?padding=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&amp;utm_source=p110">len 110</a>

**120 chars** (URL length `120`):

<a href="https://example.com/a?padding=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&amp;utm_source=p120">len 120</a>

**130 chars** (URL length `130`):

<a href="``https://example.com/a?padding=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa``&amp;utm_source=p130">len 130</a>

**140 chars** (URL length `140`):

<a href="``https://example.com/a?padding=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa``&amp;utm_source=p140">len 140</a>

**150 chars** (URL length `150`):

<a href="``https://example.com/a?padding=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&amp;utm_source=p150"&gt;&lt;img`` alt="len 150" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg" height="36"></a>

**170 chars** (URL length `170`):

<a href="``https://example.com/a?padding=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&amp;utm_source=p170"&gt;&lt;img`` alt="len 170" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg" height="36"></a>

---

mthines:
**Write-path probe (iteration 3)** — pinning the cap. Rows 1–5 bisect the segment length; the last row asks whether a *long total* URL survives when every `&`-separated segment stays short. Disposable.

first `&`-segment = **105** — total `105`, segments `105`:

<a href="``https://example.com/a?p=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa``">``&lt;img alt="first &-segment = 105" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg" height="36"&gt;``</a>

first `&`-segment = **108** — total `108`, segments `108`:

<a href="``https://example.com/a?p=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa``">``&lt;img alt="first &-segment = 108" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg" height="36"&gt;``</a>

first `&`-segment = **110** — total `110`, segments `110`:

<a href="``https://example.com/a?p=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa``">``&lt;img alt="first &-segment = 110" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg" height="36"&gt;``</a>

first `&`-segment = **112** — total `112`, segments `112`:

<a href="``https://example.com/a?p=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa``">``&lt;img alt="first &-segment = 112" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg" height="36"&gt;``</a>

first `&`-segment = **114** — total `114`, segments `114`:

<a href="``https://example.com/a?p=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa``">``&lt;img alt="first &-segment = 114" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg" height="36"&gt;``</a>

**all segments ≤ 100**, total URL **382** — segments `100, 93, 93, 93`:

<a href="``https://example.com/a?p=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&amp;q0=bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb&amp;q1=bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb&amp;q2=bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"&gt;&lt;img`` alt="all segments 100 or less, total URL 382" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg" height="36"></a>

---

mthines:
**Write-path probe (iteration 4)** — no repeated-character runs anywhere (the previous rows' `aaa…` padding was a confound). Prose-encoded prompts at increasing lengths, which is the shape the real link actually has. Disposable.

J — the proposed short link — total `132`, longest `&`-segment `73`:

<a href="https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=fix%20the%20finding%20at%20comment%20spine%20line%20review&amp;s=j">J</a>

K — total `159`, longest `&`-segment `100`:

<a href="``https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=fix%20the%20finding%20at%20comment%20spine%20line%20review%20branch%20commit%20report&amp;s=k"&gt;&lt;img`` alt="K" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg" height="36"></a>

L — total `192`, longest `&`-segment `133`:

<a href="``https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=fix%20the%20finding%20at%20comment%20spine%20line%20review%20branch%20commit%20report%20render%20guard%20check%20shape&amp;s=l"&gt;&lt;img`` alt="L" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg" height="36"></a>

M — about today's length — total `223`, longest `&`-segment `164`:

<a href="``https://app.dash0-dev.com/goto/agent0?auto_submit=true&amp;initial_prompt=fix%20the%20finding%20at%20comment%20spine%20line%20review%20branch%20commit%20report%20render%20guard%20check%20shape%20inline%20body%20write%20path&amp;s=m"&gt;&lt;img`` alt="M" src="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg" height="36"></a>

---

mthines:
issue (high): 🟠 **Relay rewrites the fix button's href** **(blocking)**

A relayed write wraps any URL over 132 chars in a code span, which closes the `href` and escapes the markup after it, so the button posts as literal text with a dead link.

Evidence: `agents/pr-reviewer/scripts/comment-spine.mjs:226` (budget)

<a href="https://app.dash0.com/goto/agent0?auto_submit=true&amp;initial_prompt=Fix%20comment-spine.mjs%3A180&amp;utm_source=pr-reviewer-fix-this"><picture><source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/mthines/agent-skills/main/agents/pr-reviewer/assets/fix-this-agent0-dark.svg">Fix with Agent0</picture></a>

<sup>`pr-reviewer` · commit `2acfa9f` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

mthines:
issue (high): 🟠 **Relay rewrites the fix button's href** **(blocking)**

A relayed write wraps any URL over 132 chars in a code span, which closes the `href` and escapes the markup after it, so the button posts as literal text with a dead link.

Evidence: `agents/pr-reviewer/scripts/comment-spine.mjs:226` (budget)

<sup>`pr-reviewer` · commit `2acfa9f` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

agents/pr-reviewer/scripts/comment-spine.mjs
```diff
@@ -0,0 +1,532 @@
...
+// total is ~20 chars above where it was rather than 60 — the title replaces the opening clause that
+// used to do its job badly, it does not add to it.
+export const TITLE_MAX = 60;
+export const PROSE_MAX = 200;
+/**
+ * The `(unverified: …)` tag on line 1. Short because line 1 is the only part of a comment that is
+ * always read, and because the agent body's own 320-char line-1 ceiling budgets for
+ * `title + decoration + prose` and nothing else.
+ */
+export const UNVERIFIED_MAX = 40;
```
mthines, line 54: issue (high): 🟠 **Maximal legal payload overflows the post pre-flight** **(blocking)**

A `suggestion` with a 60-char title, a 38-char reason and a 192-char body renders 353 chars of prose against the 320-char ceiling, and that predicate aborts the whole post rather than one comment.

Evidence: `agents/pr-reviewer.md:2889` (the 320-char ceiling) · `agents/pr-reviewer/scripts/render-comment.mjs:179` (cap set for it)

```py
if len(_prose) > 380:   # 60 title + ~25 decoration + 55 unverified tag + 200 prose
    return (False, f"comment prose > 380 chars: {len(_prose)}")
```

<sup>`pr-reviewer` · commit `59d96d2` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

agents/pr-reviewer/scripts/render-report.mjs
```diff
@@ -532,25 +556,55 @@ function main() {
...
+  // One array, three renderings. It replaces a hand-written headline, a hand-written TIER_TALLY, and
+  // nothing at all — because the report never listed what it posted inline. That absence is why the
+  // headline counted GATES (`1 error, 2 warnings`) while the inline comments were FINDINGS with no
+  // line reconciling them, and why a posted finding resorted to "the report lists all four unwired
+  // registries" — a cross-reference to a document that did not, in fact, list them.
+  const findings = arr("FINDINGS");
+  const findingRows = findings.map((f, i) => {
+    const where = `FINDINGS[${i}]`;
+    if (!f.title || String(f.title).trim() === "") fail(`${where}.title is required`);
+    assertPlain(`${where}.title`, f.title, { allowCode: true });
```
mthines, line 571: suggestion (medium): 🟡 **Report accepts a title the inline renderer rejects** **(non-blocking)**

`FINDINGS[].title` skips the 60-char cap and the noun-phrase rule, so a 198-char sentence renders into the index while the same string is rejected for the comment that shares it.

Evidence: `agents/pr-reviewer/scripts/render-comment.mjs:133` (the cap) · `agents/pr-reviewer/scripts/comment-spine.mjs:47` (shared constant)

```js
if (String(f.title).length > TITLE_MAX) fail(`${where}.title is over the cap`);
if (sentenceCount(f.title) > 0) fail(`${where}.title is a sentence, not a noun phrase`);
```

<sup>`pr-reviewer` · commit `59d96d2` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

agents/pr-reviewer/scripts/comment-spine.mjs
```diff
@@ -0,0 +1,532 @@
...
+ * The URLs in `body` that a relayed write would rewrite. Empty ⇒ the body is relay-safe.
+ *
+ * Advisory, unlike `assertPostable`: it predicts damage a specific write path would do, and the
+ * file-based path does none of it. The caller decides — the reviewer withholds fix links; nothing
+ * fails a render.
+ */
+export function relayUnsafeUrls(body) {
+  const bare = String(body).replace(/^```[\s\S]*?^```/gm, "");
+  const urls = bare.match(/https?:\/\/[^\s"'()<>]+/g) ?? [];
+  return [...new Set(urls.filter((u) => u.length > RELAY_SAFE_URL_MAX))];
```
mthines, line 308: issue (high): 🟠 **Relay check flags urls its remedy cannot remove** **(blocking)**

The documented 169-char memory deep link trips the check, but `--no-fix-links` removes only the buttons, so the re-rendered body fails identically and the gate has no exit.

Evidence: `agents/shared/rules/comment-relevance-memory.md:500` (169-char link) · `agents/pr-reviewer.md:2679` (the remedy)

```js
// Only the fix-button hrefs are withholdable, so only they are reportable.
const hrefs = bare.match(/href="(https:\/\/app\.dash0(?:-dev)?\.com\/[^"]*)"/g) ?? [];
return [...new Set(hrefs.filter((h) => h.length > RELAY_SAFE_URL_MAX))];
```

<sup>`pr-reviewer` · commit `59d96d2` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

mthines (commented):
Reviewed `59d96d2` — 3 findings — 2 blocking. 3 finding(s) inline. Sticky exists but this access path cannot edit an issue comment — report not updated in place.

---

agents/pr-reviewer/scripts/comment-spine.mjs
mthines: Confirmed and fixed in `6528909`. I measured a payload legal under every per-field cap at **363** chars of prose against the 320 ceiling — the `(unverified: …)` tag was the entire overflow; without it the same payload is 308.

Took the other branch of your suggestion rather than raising the number to 380. `payload_is_safe` already strips the fence, the button line, the evidence line, the footer and the pseudo-code note — every *rendered decoration*, measuring argument only. The unverified tag is one of those, and `UNVERIFIED_MAX` in the renderer already bounds it, so a second bound here would be the same stale copy that caused this: the cap was added to the spine and the ceiling did not move.

The reason a raised number wasn't enough on its own is that nothing tied the two together. `G46l` now renders the maximal legal payload for each shape and applies the pre-flight's own documented strips, reading the ceiling out of `payload_is_safe` with a regex — so it catches the class rather than this instance. Probed by raising `PROSE_MAX` to 320, which fails it with `429 > 320` on all three shapes.

---

agents/pr-reviewer/scripts/comment-spine.mjs
mthines: Confirmed and fixed in `6528909`, though I diverged from the suggested patch on one point.

Filtering `relayUnsafeUrls` down to fix-button hrefs makes the gate exitable, but it also throws away true information: that 169-char memory link **will** still be mangled by the relay, and a check that stops reporting it means nobody learns the citation arrived broken. So the function keeps reporting every over-budget URL, and the partition moved into a second export, `relayUnsafeFixLinks()`, which the CLI uses to pick the exit code:

- **exit 1** — a fix link is over budget. `--no-fix-links` reaches it; withhold and re-render.
- **exit 3** — something else is. The remedy cannot reach it, so post as rendered and name the mangled link in the run line.
- **exit 0** — clean.

That mirrors `--assets-check`'s vocabulary, where 1 also means withhold and 3 also means post-as-rendered, so the two checks read the same way at a call site. Both call sites now branch on the code with a `case`, and the reason is written down: `|| withhold` collapses 3 into 1 and re-renders a body that fails identically — the exact dead end you found.

Verified all three arms (`3 / 1 / 0`), plus the mixed case: a body carrying both kinds exits 1, because a remediable URL anywhere makes the run remediable. Four `G46j` guards, each probed — collapsing the partition back to `if (true)` fails two of them.

---

agents/pr-reviewer/scripts/render-report.mjs
mthines: Fixed across `6528909` and `65b21a4` — and you were right to list two checks, not one.

`6528909` added only the length cap, which left the same asymmetry one layer down: a title carrying sentence punctuation or block structure still rendered a report row and then failed the comment that row indexes. `65b21a4` completes it — `render-report.mjs` now applies all three checks `render-comment.mjs` applies to the same string (`TITLE_MAX`, `assertNoStructure`, `sentenceCount`), every one imported from the spine rather than re-derived, since a local copy of the cap is precisely the drift the spine exists to prevent.

Four `G46m` guards: the cap is imported (probed by replacing it with a local `const TITLE_MAX = 60`), and each of the three checks rejects its own violation (probed by disabling each in turn). The report's index and the finding are one string stated twice, so a bound now binds both or neither.

---

agents/pr-reviewer/scripts/render-report.mjs
```diff
@@ -835,24 +889,120 @@ function main() {
...
+    });
+    // Two phrases plus a count, never a paragraph. The cap is the same one the prose spec set at
+    // ~140 chars, expressed as a list bound so it cannot be exceeded by wording.
+    return v.length <= 2 ? v.join("; ") : `${v.slice(0, 2).join("; ")}; +${v.length - 2} more`;
+  };
+  if (verdict === "FAIL" && arr("FAIL_REASONS").length === 0) {
+    fail("VERDICT FAIL with no FAIL_REASONS — a failing gate names why in one noun phrase"
+      + " (report-rendering.md § Headlines)");
+  }
+  if (arr("FAIL_REASONS").length > failing) {
```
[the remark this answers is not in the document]
mthines: Confirmed and fixed in `9ae09c3`.

Worth recording *why* this went unnoticed, because it is the reason it deserved more than a nitpick: `reasonList` truncates the rendering at two phrases plus `+N more`, so an over-long list never **looked** wrong. A `WARN` claiming five warnings above a table showing one ⚠️ silently dropped the third phrase rather than displaying a contradiction — the same class this renderer exists to make unrepresentable (a `VERDICT` that disagrees with its gate table, a `FINDINGS` length that disagrees with its own count).

Both polarities now, plus the empty-list requirement the `FAIL` branch already had and `WARN` did not. One asymmetry from the `FAIL` branch is deliberate: CI *is* countable here, since it is a warning gate and `warning` already includes it, and the message says `(CI included)` when `CI_NOTE` is set.

Three `G46m` guards, each probed by reverting the check it pins — including one that asserts the symmetry in the *source* rather than only in behaviour, so removing either arm fails.

The same-file title finding above turned out to be this exact shape, which your comment called out independently. Both are closed.

---

agents/pr-reviewer/scripts/comment-spine.mjs
```diff
@@ -0,0 +1,564 @@
...
+ * `<source>` with the light asset doing double duty as the default. `<picture>` picks a `<source>`
+ * by media query and, crucially, does NOT fall back to `<img>` when the chosen resource fails to
+ * load, so the default is not a safety net for a bad `srcset`; it is what renderers that ignore
+ * `<source>` altogether display — and that set includes GitHub's own notification emails and RSS,
+ * where an inline finding is very often first read. Pointing the default at the theme-suffixed
+ * light file made those readers depend on a file that only shipped with the theme split; the
+ * unsuffixed `{stem}.svg` has been on the default branch since before it, so it degrades to a
+ * button rather than to a broken image.
+ */
+export function fixButton({ kind, url, label }) {
```
mthines, line 170: issue (high): 🟠 **Fix button costs 1225 chars against a 2000-char ceiling** **(blocking)**

The budget above that check allows ~430, so an ordinary two-evidence finding renders 2049 chars and `payload_is_safe` aborts the whole post rather than one comment. A maximal payload reaches 2699.

Evidence: `agents/pr-reviewer.md:2915` (budgets the button at ~430) · `agents/pr-reviewer.md:2917` (the 2000 ceiling)

```py
# strip the button line first, as the prose check above already does
_body = _re.sub(r'^<a href="https://app\.dash0(?:-dev)?\.com/.*$', "",
                c.get("body", ""), flags=_re.MULTILINE)
if len(_body) > 2000:
```

<sup>`pr-reviewer` · commit `65b21a4` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

agents/pr-reviewer/scripts/render-comment.mjs
```diff
@@ -0,0 +1,520 @@
...
+    // `render-report.mjs` was rejected as markup, and the bare form trips `sentenceCount` on the
+    // dot. The title renders inside `**…**`, where an inline code span is well-formed, and
+    // `sentenceCount` already ignores punctuation inside backticks.
+    assertPlain("TITLE", title, { allowCode: true });
+    assertNoStructure("TITLE", title);
+    if (title.length > TITLE_MAX) {
+      bad(`TITLE is ${title.length} chars, over the ${TITLE_MAX}-char cap — it is a noun phrase,`
+        + " not a sentence");
+    }
+    if (sentenceCount(title) > 0) {
```
mthines, line 137: issue (medium): 🟡 **Piped title accepted inline, rejected by the report** **(non-blocking)**

A `|` in TITLE renders here but `FINDINGS[].title` rejects it, so the report is lost after every comment has already cleared the per-finding gate. The parity work bound the other three checks.

Evidence: `agents/pr-reviewer/scripts/render-report.mjs:578` (pipe rejected) · `scripts/eval/l1.mjs:3719` (G46m pins three checks)

```js
if (title.includes("|")) bad("TITLE contains a pipe — the report row that
  indexes this finding rejects it");
```

<sup>`pr-reviewer` · commit `65b21a4` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

agents/pr-reviewer.md
```diff
@@ -2540,19 +2619,106 @@ printf '%s\n' "$REPORT_BODY" | awk '
...
+1. **Check before the write, and withhold the buttons rather than post them broken.** No amount of
+   faithful copying saves an over-budget URL:
+
+   ```bash
+   # Exit 1 is a fix link over budget — the remedy removes it. Exit 3 is some OTHER long URL
+   # (a cited doc, an asset path): --no-fix-links would not remove it, so re-rendering is a
+   # loop with no exit. Post as rendered and name the mangled link in the run line.
+   node "$AGENT_SUPPORT/pr-reviewer/scripts/comment-spine.mjs" --relay-check /tmp/report-body.md
+   case $? in
+     1) RERENDER_WITH_NO_FIX_LINKS=1 ;;   # re-render, post that, and say so in the run line
```
mthines, line 2683: suggestion (low): ⚪ **Exit-3 obligation undischarged after an exit-1 remedy** **(non-blocking)**

With both a fix link and a cited link over budget the check exits 1; after withholding, the surviving exit-3 condition is never re-checked, so the run line never names the mangled citation.

Evidence: `agents/pr-reviewer/scripts/comment-spine.mjs:331` (the partition)

```bash
# after re-rendering without fix links, ask again — 3 may survive
node "$SPINE" --relay-check /tmp/report-body.md; [ $? -eq 3 ] && note_mangled_link
```

<sup>`pr-reviewer` · commit `65b21a4` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

agents/pr-reviewer/scripts/comment-spine.mjs
```diff
@@ -0,0 +1,564 @@
...
+}
+
+/**
+ * Prose that must not carry markup the renderer adds itself.
+ *
+ * Two bars, and the split is deliberate. Identifier fields (`path`, `key`, `title`) are wrapped or
+ * emphasised by the renderer, so a backtick inside one would terminate that span. Prose fields keep
+ * their inline code, because a finding body naming a symbol without backticks fails its own rule.
+ */
+export function assertPlain(where, v, { allowCode = false } = {}) {
```
mthines, line 411: suggestion (low): ⚪ **Spine assertPlain has a local twin in the report renderer** **(non-blocking)**

`render-report.mjs:135` keeps a behaviourally identical copy that calls `fail()` rather than throwing, and G46f's no-local-copy check covers only `TIER_GLYPH` and `footerLine`.

Evidence: `agents/pr-reviewer/scripts/render-report.mjs:135` (the twin) · `scripts/eval/l1.mjs:3470` (G46f's two-symbol list)

```js
import { assertPlain as plain } from "./comment-spine.mjs";
const assertPlain = (w, v, o) => { try { plain(w, v, o); } catch (e) { fail(e.message); } };
```

<sup>`pr-reviewer` · commit `65b21a4` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

mthines (commented):
Reviewed `65b21a4` — 🟠 4 findings — 1 blocking 4 finding(s) inline. Sticky exists but this access path cannot edit an issue comment — report not updated in place.

---

agents/pr-reviewer/scripts/comment-spine.mjs
mthines: Confirmed and fixed in `0a19869`. I measured it before fixing it: a legal `issue:` at every cap — 60-char title, 200 chars of prose, the fixture's two evidence refs, a 10-line fence and a button on a modest 408-char link — renders **2208** chars, of which **933** is the button. Over a ceiling that aborts the whole post.

Took your patch as written, and the reason not to just raise 2000 is the one this PR already hit once: `build-agent0-link.mjs` bounds the URL at `MAX_URL = 4000`, so a maximal button is ~4525 chars on its own and no fixed body ceiling can leave room for it. The button belongs to the same class as the fence, the evidence line and the `(unverified: …)` tag — a rendered affordance, not argument — and every one of those is already stripped before the prose measurement.

The `~430 chars` in the adjacent comment was measured before the theme split turned one `<source>` into two plus an ``; it now carries the real number and the reason it will drift again.

What made this reachable at all is that the two ceilings in one predicate had asymmetric coverage: `G46l` measured prose and nothing measured the body. Both are now read out of the source and both are measured against a maximal render, including a new shape carrying a 10-line fence **and** a button. One detail worth naming — my first version of that guard hand-copied the strip, so removing the strip from `payload_is_safe` failed only the source-shape check and the measurement still passed. It now applies the strip only if the source does, which reproduces your exact defect as `2245 > 2000`.

Six probes across the four findings, each failing only its own check.

---

agents/pr-reviewer/scripts/render-comment.mjs
mthines: Confirmed and fixed in `0a19869`, and you're right that the parity work bound the other three and stopped one short.

The failure ordering is what makes this worse than a cosmetic gap: the inline renderer runs per finding and the report renderer runs once, after, so a piped title clears every per-finding gate, posts every comment, and *then* loses the whole report — the surface that carries the index and the gate table. The three checks that already had parity all fail at the same point on both surfaces; this one failed at different points.

`render-comment.mjs` now rejects it with the reason stated in terms of the other surface (a table cell cannot be escaped out of), and G46f pins it three ways: both renderers reject a pipe in that string, the inline renderer rejects a real piped payload, and the rejection emits nothing on stdout. Probed by disabling the check — which, usefully, printed the offending title (`**Fix button | costs too much**`) as the guard's own failure message, so the defect is visible in the probe output rather than only argued.

---

agents/pr-reviewer.md
mthines: Confirmed and fixed in `0a19869`. The premise is exactly right and it is a consequence I wrote into the partition without following through: exit 1 wins over 3 because *a remediable URL anywhere makes the run remediable* — which says something about the run, not that the body is otherwise clean. A body carrying both kinds therefore re-rendered, posted, and never named the citation the relay was still going to mangle, on precisely the runs that had two problems rather than one.

Fixed at all three places that state the contract, since discharging it in one and dropping it in the others is how this happened in the first place: the inline call site, the sticky call site, and the exit-code table in `agent0-fix-links.md` that owns the vocabulary. The sticky site also gained the `NOTE_MANGLED_LINK` variable it was missing — the `3) : ;;` arm was a no-op, so even the *first* exit-3 had nowhere to go.

Four `G46j` guards: each of the three sites re-checks after the remedy (probed by deleting the re-check at each in turn — each fails only its own), plus one asserting the result actually reaches the run line, since a re-check whose answer goes nowhere is the same silence with an extra step. That last one is probed by replacing `NOTE_MANGLED_LINK=1` with `true`.

---

agents/pr-reviewer/scripts/comment-spine.mjs
mthines: Confirmed and fixed in `0a19869`. This one is the change indicting itself: the same commit that imported `assertNoStructure` and `sentenceCount` from the spine walked past the third check on those same fields and left it local. So the two surfaces went on validating identical fields against two definitions of "plain" — inside the change whose whole thesis is that they must not.

`render-report.mjs` now wraps `spineAssertPlain`, and the wrapper is the entire remaining difference: the spine throws, and this renderer must exit non-zero with nothing on stdout so a rejection can never post a partial body. The report-specific half of the old docblock is kept — why prose fields need `allowCode: true` (Step 1.0 requires an `ask` be another bot's lead line truncated rather than paraphrased, and those lead lines carry backticks) — because that rationale is about this caller, not about the check.

On G46f's two-symbol list: I didn't extend it, because the generic test asserts *no local binding*, and a delegating wrapper is a local binding. Extending it would have forced a rename of every call site to satisfy a guard whose real target is a second **implementation**. So `assertPlain` gets its own form, detecting a reimplementation by the spine's own error strings (`contains a markdown link`, `contains a backtick`) appearing in a renderer — which is what a copy actually looks like, and which the twin you found would have tripped from the day it was written. Probed by pasting a condensed copy back in.

---

PR #166, opened 2026-09-04
Fix buttons invoke Agent0's /implement skill instead of carrying the method

mthines:
The two "Fix with Agent0" buttons carried the whole fix method inside their URL.
**Fix all** embedded a `gh api graphql` worklist query — a `pageInfo`/`endCursor` walk plus a client-side login filter — at ~1100 encoded characters, and **Fix this** embedded the finding's own lead line at ~880.
Agent0's `/implement` skill owns that method now: gather the PR's comments, filter them to one author, apply the actionable ones, commit, and push.
So the prompt only has to name the target.

```text
/implement https://github.com/{owner}/{repo}/pull/{n} {bot_login}
/implement https://github.com/{owner}/{repo}/pull/{n} {bot_login} — apply only the comment at {path}:{line}.
```

That is ~190 characters and flat for **Fix all**, and ~360 at its worst case for **Fix this**.

Three things the prompts stop carrying, each now the skill's: the embedded query, which was a second implementation of gathering that drifts; the `{lead}` excerpt, which went stale on a `§ Hard caps` prose trim and is what doubled **Fix this**; and the `{count}` checksum, since `/implement` reports what it applied.

The `<author>` argument is always filled, never left to `/implement`'s own resolution — it excludes bot authors unless one is named, so an omitted login would silently skip every finding of a reviewer posting as a bot.
It is also what keeps the run off every other author's comments, which is the safety property the retired query's client-side filter provided.

Two constraints are worth calling out.
**Fix this** cannot carry its own inline-comment permalink: GitHub assigns a review comment's `r<id>` only on POST, and inline comments here are append-only, so a comment cannot link to itself — hence `{path}:{line}`.
**Fix all** gains one: when the login is unresolvable but a prior sticky was matched, it passes the report comment's `#issuecomment-<id>` permalink instead, which narrows the omit-the-button rule rather than widening it.

One dependency, stated in the rule file so it stays auditable: the two `/implement` templates no longer carry the "repo's own scripts, never a raw `tsc`/`eslint` call, never a whole-repo pass" verification clause.
If `/implement` does not verify that way, the runner OOM crash observed in review of #151 regresses through these buttons.
The CI-only variant is not an `/implement` call — a red check with zero findings has no comments to apply — and keeps the clause inline.

Guards: `G32a`/`G32c` pin both argument grammars, `G32l` the permalink fallback, `G32m` rejects each re-inlined shape by name, `G32n` keeps the CI-only variant out of `/implement`, and `G32k` holds both URLs under 500 characters so a re-added clause reads as a length regression instead of vanishing into the headroom.
The retired embedded-query guards go with the query.
L1 is 1036 → 1039/1039, and all eight new guards were probed by breaking what they guard.

Also on this branch: `c1d6fb7` closes the `pr-reviewer` memory record schemas against invented fields, with four `G38e` guards.

https://claude.ai/code/session_01J6iWLkrnKDsk8YNsjaKQq4

---
_Generated by [Claude Code](https://claude.ai/code/session_01J6iWLkrnKDsk8YNsjaKQq4)_


mthines:
Closing in favour of [#165](https://github.com/mthines/agent-skills/pull/165), which was started from the same request in a parallel session. **This PR had the better design; that one has the larger body of work**, so the split is by branch, not by idea.

**Why #165 is the branch that survives.** This PR's 7 files are a strict subset of #165's 44. #165 carries the whole report/inline comment-system unification (`82f8ac8`), the fix-button width pin, the renderer↔write copy guard, the relay-length withhold path, and the asset-availability fix — five commits that would all have to be re-landed here. This PR is two commits, and both port cleanly the other way.

**Why this PR's prompt design wins anyway.** #165 had reached the same conclusion — the method does not belong in a URL parameter — and was about to author a new `fix-review` skill in `skills/delivery/` to host it, gated behind a new `agent0_fix_skill:` review-config field because whether a skill is installed on the Agent0 side is not knowable from the reviewing repo. Invoking Agent0's existing **`/implement`** is the better answer to the same problem: no new skill, no new config field, no installation question, and no second implementation of "gather this PR's comments and filter them to one author" to drift against the one Agent0 already has. That uncommitted `fix-review` work has been dropped from #165 rather than landed.

Two things here are strictly better than what #165 had and should come across with it:

- The **login fallback** — passing the sticky's `#issuecomment-<id>` permalink when `{bot_login}` is unresolvable. It *narrows* the omit-the-button rule instead of widening it, which #165's version did not do.
- Stating the **delegated-verification dependency** in the rule file. Moving the "repo's own scripts, never a raw `tsc`/`eslint`, never a whole-repo pass" clause into the skill makes the OOM crash from review of #151 reachable through these buttons if `/implement` does not verify that way. Auditable beats forgotten.

One caveat worth recording before the design is re-landed: `/implement`'s argument grammar, its single-author default, and its bot-author exclusion are all asserted here but not verifiable from this repo. The templates are correct only if those hold, so they want confirming against the live skill — and the per-comment scope clause on **Fix this** sits outside the two-positional grammar by construction.

**Nothing is lost by closing.** `c1d6fb7` (memory record schemas closed against invented fields) is unrelated to the fix buttons and was swept onto this branch incidentally — it has been cherry-picked onto #165 as `59d96d2`, L1 1199 → 1203, so it is not parked. The `/implement` change itself stays on `claude/pr-reviewer-quality-1kn8ba`, which is **not** being deleted: `git cherry-pick 5cd0c61` re-lands it on top of #165, with conflicts expected in `agent0-fix-links.md` and `l1.mjs` (this branch predates #165's `## Asset availability`, `## Relay length limit`, and the `on-wherever-an-Agent0-exists` opt-in default, and its diff still shows the old "off by default" text).

---

PR #168, opened 2026-09-04
Fix buttons invoke Agent0's /pr-fix skill instead of carrying the method

mthines:
The two "Fix with Agent0" buttons carried the whole method in their URL. **Fix all** embedded a `gh api graphql` worklist query with a `pageInfo` walk and a client-side login filter (~1100 encoded chars); **Fix this** embedded the finding's own lead line (~880). Agent0's `/pr-fix` skill owns that method — gather the PR's comments, filter to one author, apply, commit, push — so the prompt only has to name the target.

Along the way this PR found and fixed the reason **the buttons had never rendered at all** (see *The buttons never rendered*, below).

This re-lands the design from [#166](https://github.com/mthines/agent-skills/pull/166), closed unmerged in favour of [#165](https://github.com/mthines/agent-skills/pull/165) because #165's 44 files would all have had to be re-landed the other way. #166's closing comment recorded the recipe (`git cherry-pick 5cd0c61`, conflicts expected in `agent0-fix-links.md` and `l1.mjs`); this is that, with the skill renamed from `/implement` to `/pr-fix`.

## The templates

```text
/pr-fix https://github.com/{owner}/{repo}/pull/{n} {bot_login}
/pr-fix https://github.com/{owner}/{repo}/pull/{n} {bot_login} — apply only the comment at {path}:{line}.
```

Fix all is ~190 chars and flat in the finding count; Fix this is ~360 at its worst case (a 94-char path). The trailing clause is the one thing outside `/pr-fix`'s two-argument grammar and it earns its place: the grammar has no per-comment scope, so without it both buttons would do the same thing.

**Three things the prompts stop carrying**, each now the skill's. The embedded query — a second implementation of gathering that drifts. The `{lead}` excerpt — it went stale on a Hard-caps prose trim, and quoting a body into a URL is what doubled Fix this. The `{count}` checksum — `/pr-fix` reports what it applied.

**The author argument is always filled**, never left to `/pr-fix`'s own resolution: it excludes bot authors unless one is named, so an omitted login would silently skip every finding of a reviewer posting as a bot. It is also what keeps the run off every other author's comments.

**New: a login fallback.** When the login is unresolvable but a prior sticky was matched, the report comment's own permalink is passed instead. A comment permalink is a PR URL with a fragment, and the comment it names is the reviewer's own, so `/pr-fix` resolves the author by naming it. Fix this gets no such fallback: GitHub assigns a review comment's own id only on POST, and inline comments are append-only, so a comment cannot link to itself. That divergence is now stated in both owners — see the invariant note below.

**The CI-only variant stays prose and must not become a `/pr-fix` call** — a red check with zero findings has no comments to apply. It keeps the verification clause (repo's own scripts, never a raw `tsc`/`eslint` call, never a whole-repo pass) that the two `/pr-fix` templates delegate to the skill.

## The buttons never rendered

`--relay-check` asks whether a *relayed* write would mangle a long URL. It was being run **unconditionally**, and every fix link exceeds the 140-char budget by construction — the floor is 164 — so the withhold fired on every run of every repo, `gh` runs included, where nothing would have been rewritten. A default-on affordance withheld on all paths is indistinguishable from one that is switched off.

The condition was known and written down, which is the interesting part. The report block carried *"on the `gh` path the buttons post intact and stay"* as prose, directly beneath a shell block that asked unconditionally; the inline block had neither the condition nor the sentence. **A rule the shell does not execute is a rule the run does not follow.**

Both call sites are now gated on `ACCESS_PATH == "mcp"`, and `github-access.md § Step 0` binds `ACCESS_PATH` (`gh` | `mcp` | `none`) as a named output with a per-row column for **how the body travels** — a file, or a tool-call argument. Undecided probes resolve to `mcp`: withholding is recoverable, a mangled button is not.

The first attempt at that gate (`e8c62d0`) did not work, and the next review caught it: the probe read `$RESOLVED_REPO`, bound at one site in a different tool call, so it probed `repos/`, 404'd, and pinned `ACCESS_PATH` to `mcp` anyway. `8a20a5d` derives the repo with § Step 0's own fallback chain — the protection that rule carries and I had copied the probe without.

## Three merge resolutions against #165, none of them mechanical

- **Fix this** now passes the deep link as the payload's `FIX_URL` and lets `render-comment.mjs` build the button via `comment-spine`'s `fixButton()`. #165 moved that markup behind the shared spine, so taking its side was required, not preferred.
- The parked branch put a `FIX_ALL_URL` on the **`pass`** report fixture. #165's renderer *fails closed* on a fix button with zero findings and no CI note, so the button belongs on **`warn`**. `TIER_TALLY` went with it — #165 retired that payload key.
- The rule file keeps #165's on-by-default text; the parked branch's copy still said "off by default".

## Commits after the re-land

| Commit | What |
| --- | --- |
| `3ed08e9` | Both findings from review 1: the login fallback is gated on `OPEN_FINDING_COUNT` as well as identity (the bullets are first-match-wins, so identity alone pre-empted the CI-only variant and the omit rule), and `CLAUDE.md`'s "always named" absolute reworded so the permalink fallback satisfies it. |
| `aafa7da` | The inline footer carries the identity half only — agent name and commit. The methodology link is report-only: one review, one link, on the object that is the review. `footerLine()` gains a `docs` flag defaulting **off**, so the shared builder stays shared (`G46f`) and a surface has to ask. Guarded both directions — `G46e` absence inline, `G25` presence on the report. |
| `50db0d7` | Review 2: the both-buttons-or-neither invariant scoped to the flag, with the one identity state that diverges the placements named in both owners (`G32o`); the relay figure made re-derivable by shipping an executable recipe instead of a cited number. |
| `e8c62d0` | The write-path gate (above), plus `ACCESS_PATH` as a named contract. |
| `8a20a5d` | Review 3: the probe now resolves; the exit-3 re-check re-renders before re-asking (it was re-reading an unchanged file, so `NOTE_MANGLED_LINK` was unreachable); a guard message that named the opposite fail-safe direction; one re-wrap. |

## What a reviewer should push back on

`/pr-fix`'s argument grammar, its single-author default, its bot-author exclusion, and its verification behaviour are **asserted by these templates and not verifiable from this repo**. The templates are correct only if those hold. The verification dependency is the one with teeth and is stated in the rule file so it stays auditable: if `/pr-fix` ever verifies with a raw whole-graph `tsc`, the runner crash observed on review of #151 regresses through these buttons.

The rename itself is also unconfirmed from here — nothing in this repo establishes that a skill named `/pr-fix` exists on the Agent0 side.

## Verification

- `node scripts/eval/l1.mjs` — 1256 → **1279, all green**. `render-comment.mjs --self-test` 40/40. `validate-report-shape.mjs` clean on all four committed report bodies.
- New guards: `G32l` (the fallback's count gate), `G32o` (the invariant's flag scope, both owners), `G32p` (the write-path gate **and** that the probe can actually reach `gh`, plus the exit-3 re-render, asserted positionally). `G32k` still holds both URLs under 500 chars.
- **Every guard probed by breaking what it protects**, including the three added after each review round. `G32p`'s first version passed while the feature stayed broken — it asserted the gate existed rather than that it worked — which is why it now measures the probe's reachability.
- The relay figures are re-derivable rather than remembered: the recipe is in the rule file and reproduces 189 / 197 / 200 exactly, the login being the only variable (+4 on `development`).

### The relay length limit, confirmed by observation

`RELAY_SAFE_URL_MAX = 140` was derived; it is now measured on a live write path that carries the body as a tool-call argument. Four variants posted and read back:

| Variant | URL length | Result |
| --- | --- | --- |
| Rendered Fix-all button, full deep link | 195 | **destroyed** — a code span opens inside the anchor's URL attribute and closes mid-markup, escaping the elements after it |
| Same anchor and picture markup, 44-char URL | 44 | **untouched, byte-for-byte** |
| Long deep link as a plain markdown link, no HTML | 195 | **destroyed** |
| Long deep link bare, no link syntax | 195 | wrapped in a code span; renders as code, not a link |

The discriminator is **URL length, not the markup**. So the floor-above-ceiling arithmetic is the whole story: scaffold 105 body chars empty, shortest conceivable filled link 164, realistic fills 189–204, budget 140. No prompt fits, which is why *"do not shorten the prompt to fit"* is an absolute. The remedy is the write path — post the body from a file (`gh api … --field body=@file`, `gh pr review --body-file`) and no rewrite happens.

https://claude.ai/code/session_015z8UFRtJGJXoUd66beoNRz

---
_Generated by [Claude Code](https://claude.ai/code/session_015z8UFRtJGJXoUd66beoNRz)_


mthines:
### 🟡 2 findings

Retires the embedded worklist query, the `{lead}` excerpt and the `{count}` checksum from both Agent0 fix prompts into a `/pr-fix` call plus an address.

**Warnings:** 2 non-blocking findings

|---|---|---|
| Login-fallback branch is not gated on the finding count | `agents/pr-reviewer.md:369` | 🟡 medium |
| “always filled” has an exception added in this PR | `CLAUDE.md:110` | ⚪ low |

<details>

**Needs attention**

| Gate | Status | Details |
|---|---|---|
| Description vs. code | ✅ | The description matches the diff, including the figures and guard names, which were re-measured. |
| Prior review feedback | ✅ | No open review threads. |
| Documentation | ✅ | The rewrite states its own unverifiable dependency on `/pr-fix` rather than eliding it. |
| Self-review signals | ✅ | No debug logs, leftover markers, or unreviewed stubs. |
| Code review | ⚠️ | 2 non-blocking findings — see inline comments. |

<details>
<summary>Withheld (1) — could not be verified from this runner</summary>

- `agents/shared/rules/agent0-fix-links.md:180` — question: The trailing `— apply only the comment at {path}:{line}.` clause may not parse as `/pr-fix`'s two positional arguments <sup>(unverified: the `/pr-fix` skill is not in this repository, so its argument parser cannot be read)</sup>

</details>

**Found**

Quality — produced 4 → posted inline 2 · cleared 2 · carried forward 0 · deferred 0 · below-bar 0
Severity — 🟡 1 medium · ⚪ 1 low
Standards — ran · 1 doc · 1 finding(s)
Verified — `node scripts/eval/l1.mjs` — 1259/1259 green. Both templates re-measured from the live rule file: Fix-all 188 chars, Fix-this 353 worst case, fallback 193 — all under the 500-char `G32k` bound.

**Run**

full · 421 lines in delta · tier deep · depth checkout · 8 files touched; blast_radius=none, semver_delta=none, high_stakes=0, risky_shapes=none
⚠️ Memory fan-out narrowed to `repo::` scope only — the two `global` reviewer-lesson/relevance pages were not read this run.
CI — `l1` and `l2` both green.
Memories — 0 indexed · 0 used

<sup>Nothing to report — optimality (2 judged), measurability (skipped), integrations (not activated), 0 files skipped.</sup>

</details>

<sup>`pr-reviewer` · commit `1b86482` · full review · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md) · updated 2026-09-04 08:05 UTC</sup>

---

agents/pr-reviewer.md
```diff
@@ -363,13 +363,13 @@ Resolve `AGENT0_FIX_LINKS` and `AGENT0_ENVIRONMENT` per `review-config.md § Run
...
+  - **When `{bot_login}` is resolved and `OPEN_FINDING_COUNT` is non-zero**, use the `/pr-fix` prompt from `agent0-fix-links.md § Prompt templates`, filled with the PR URL and `{bot_login}`. Fill the `<author>` argument always — `/pr-fix` excludes bot authors unless one is named, so an omitted login silently skips every finding of a reviewer posting as a bot.
+  - **When `{bot_login}` is unresolved** (both `ME` and `PRIOR_REPORT_AUTHOR` empty) **but a prior sticky was matched**, use the login-fallback template instead, filled with `https://github.com/OWNER/REPO/pull/<n>#issuecomment-<sticky_id>` — the id the Step 0.7 marker match already returned for the PATCH, never a fresh lookup. Naming the reviewer's own comment is how `/pr-fix` resolves the author without a login.
```
mthines: issue (medium): 🟡 **Login-fallback branch is not gated on the finding count** **(non-blocking)**

Its condition is identity-only, so an unresolved login with a matched sticky pre-empts the CI-only variant and the omit rule below. The old `regardless of {count} or CI state` marker is gone.

Evidence: `agents/pr-reviewer.md:370` (CI-only branch) · `agents/shared/rules/agent0-fix-links.md:230` (omit rule)

```text
- **When `{bot_login}` is unresolved** … **but a prior sticky was matched**
  **and `OPEN_FINDING_COUNT` is non-zero**, use the login-fallback template…
```

<sup>`pr-reviewer` · commit `1b86482` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

CLAUDE.md
```diff
@@ -107,7 +107,7 @@ They are **generated from templates**, not stored as `agents/*.md`, so searching
...
...(for *Fix this*) the one `path:line` to scope to — never the method: the embedded `gh api graphql` worklist query, the `{lead}` excerpt, and the `{count}` checksum were all retired into Agent0's own `/pr-fix` skill, taking *Fix all* from ~1100 encoded chars to ~190 (flat in the finding count) and *Fix this* from ~880 to ~360, with `G32k` holding both under a 500-char regression bound so a re-added clause reads as a regression instead of vanishing into headroom. Two consequences are load-bearing: the `<author>` argument is **always** filled (`/pr-fix` excludes bot authors unless one is named, so an omitted login would silently skip every finding of a reviewer posting as a bot, and it is also what keeps the run off every other author's comments), and because the prompt no longer quotes a body, the link is stable across a prose trim. `/pr-fix`'s argument grammar, its single-author default, and its verification behaviour are asserted here but **not verifiable from this repo** — the templates are correct only if those hold, and the delegated-verification dependency (repo's own scripts, never a raw `tsc`/`eslint`, never a whole-repo pass) is stated in the rule file so it stays auditable.
```
mthines: suggestion (low): ⚪ **“always filled” has an exception added in this PR** **(non-blocking)**

The new login fallback `/pr-fix {report_comment_url}` fills no `<author>` argument, so this absolute reads as unqualified while `agent0-fix-links.md` documents the exception.

Evidence: `agents/shared/rules/agent0-fix-links.md:218` (fallback template)

```text
the `<author>` argument is **always** filled *except in the login fallback,
where the report comment's own permalink names the author instead*
```

<sup>`pr-reviewer` · commit `1b86482` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md)</sup>



---

mthines (commented):
---

agents/pr-reviewer.md
mthines: Confirmed and fixed in `3ed08e9`. The diagnosis was right down to the cause — the branch this replaced carried `regardless of {count} or CI state` as an explicit precedence marker, and the rewrite dropped the phrase, which is what turned a terse bullet into an ambiguous one.

The bullet now reads `**a prior sticky was matched**, **and `OPEN_FINDING_COUNT` is non-zero**`, plus a sentence saying why the count condition is not redundant with the bullet above it: identity picks *which* `/pr-fix` template, the count decides whether a `/pr-fix` template is the right one at all.

`agent0-fix-links.md:230` is fixed too. Its `omit … only when both are unavailable` was the same error stated as prose — it reads as licensing a fallback at a count of 0 — so the two owners had drifted and only the rule file was wrong. It now states identity and count as independent conditions, with the count restated on the fallback form itself.

Guarded so it cannot regress: `G32l` gains two checks. The first anchors on this bullet's own sentence rather than counting `OPEN_FINDING_COUNT` across the file — the identifier appears in four bullets here, so a whole-body match would have passed with this exact condition deleted. Both probed by breaking them: reverting the bullet to identity-only reproduces the defect you found, deleting the rule-file paragraph fails the second. L1 1259 → 1261.

---

CLAUDE.md
mthines: Fixed in `3ed08e9`.

The finding was right that the absolute did not survive its own next sentence. `CLAUDE.md` said the author is "always named, never left to `/pr-fix`'s own resolution", and the login-fallback template names nobody — it passes the sticky's `#issuecomment-<id>` permalink as the `<pr>` slot and lets the comment it points at resolve the author. Two mechanisms, and the sentence described one.

The reword keeps the absolute and makes the fallback the second way of satisfying it, rather than an exception to it:

> the author is **always named, never left to `/pr-fix`'s own resolution** — either as the `<author>` argument (…) or, when the identity ladder resolves nothing and a prior sticky was matched, by passing that sticky's own `#issuecomment-<id>` permalink as the `<pr>` slot so the comment it names resolves the author with no `<author>` argument at all

That is accurate: in both branches the author is determined by something the link carries, and never by `/pr-fix`'s default author selection. `README.md` carries the shorter form of the same clause.

---

agents/pr-reviewer.md
```diff
@@ -363,13 +363,13 @@ Resolve `AGENT0_FIX_LINKS` and `AGENT0_ENVIRONMENT` per `review-config.md § Run
...
...p link as the payload's `FIX_URL` and let `render-comment.mjs` build the button (theme-aware markup, host validation and alt text all live in `comment-spine.mjs`'s `fixButton()`; never hand-write the markup), built with **the same `--env <env>` resolved above** — `node "$BUILD_LINK" --env <env> --source fix-this "<fix-this prompt>"` (`--source` is mandatory and is `fix-this` here, always — `agent0-fix-links.md § Click attribution`; never the bare relative path — see the Fix-all bullet's note; a bare path silently resolves against whatever the shell's cwd is, which during a cross-repo dispatch is the *reviewed* repo, not this one) — and the fix-this prompt template, filled with the PR URL, `{bot_login}`, and this finding's own `path:line`. The finding's **body plays no part in the URL** — no lead line, no quoted text — so the link is stable across a § Hard caps prose trim. Skip the button when `{bot_login}` is unresolved (the inline template has no comment-permalink fallback: a comment cannot link to itself), and skip it for `nitpick` / `question` / `praise`. **This is the same `<env>` as the Fix-all bullet above, resolved once per run — never re-resolved or defaulted per finding.**
```
mthines, line 372: issue (medium): 🟡 **Login fallback breaks the both-buttons-or-neither rule** **(non-blocking)**

With `ME` and `PRIOR_REPORT_AUTHOR` both empty but a sticky matched, line 369 renders Fix all while this line skips every Fix this.

Evidence: `agents/pr-reviewer.md:376` (both or neither) · `agents/shared/rules/agent0-fix-links.md:67` (same claim) · `agents/pr-reviewer.md:3181` (empty bot_login persists)

<sup>`pr-reviewer` · commit `20cf328`</sup>



---

agents/shared/rules/agent0-fix-links.md
```diff
@@ -439,6 +489,21 @@ design the link around. A `fix-this` link spends ~110 body chars before the prom
 ~30 — enough for `Fix <basename>:<line>` and not the repo, the PR, or the finding. A button that
 opens a session with no idea what to fix is worse than no button.
 
+**On a relayed write the buttons are unreachable by construction — the floor is above the ceiling.**
+The deep-link scaffold alone (`https://app.dash0.com/goto/agent0?auto_submit=true&initial_prompt=`
+plus `&utm_source=pr-reviewer-fix-all`) is **105 body chars** empty, leaving 35 for the encoded
+prompt; the shortest conceivable filled Fix-all — a one-character owner, repo and login, PR #1 —
+measures **164**, and a realistic one (`mthines/agent-skills` #168) measures **195**. Every one is
```
mthines: issue (low): ⚪ **Relay figure 195 does not reproduce for the PR it cites** **(non-blocking)**

Filled for `mthines/agent-skills#168` the Fix-all link measures 189 in-body, 49 over 140. 195/55 is `lorekit#594` with a bot login.

Evidence: `agents/shared/rules/agent0-fix-links.md:501` (the 55 shortfall) · `agents/pr-reviewer/scripts/build-agent0-link.mjs:1` (measured via buildLink)

```text
a realistic one (`mthines/agent-skills` #168) measures **189**.
... which those figures still exceed by 49 chars.
```

<sup>`pr-reviewer` · commit `20cf328`</sup>



---

agents/shared/rules/agent0-fix-links.md
```diff
@@ -439,6 +489,21 @@ design the link around. A `fix-this` link spends ~110 body chars before the prom
...
+over 140 before any prompt content is chosen. This is why *"do not shorten the prompt to fit"* is stated as an absolute rather
```
mthines: nitpick (low): This line is 126 chars where the paragraph wraps near 100 — the edit merged two sentences without re-wrapping. **(non-blocking)**

<sup>`pr-reviewer` · commit `20cf328`</sup>

---

mthines (commented):
Reviewed `20cf328` — 3 findings — 1 medium, 2 low, none blocking. 3 finding(s) inline. Sticky exists but this access path cannot edit an issue comment — report not updated in place.

---

agents/pr-reviewer.md
mthines: Fixed in `50db0d7`, and the reachability trace through the state write at line 3181 is what made it worth fixing rather than arguing with.

The invariant was false exactly where a reader would go looking, and the population argument is the part that lands: a run reaching the fallback bullet has one button **every time**, not occasionally, so "both or neither" was not an approximation with a rare corner — it was wrong for the whole branch.

What it is *not* is a missing button to be supplied. Fix this has no comment-permalink fallback because GitHub assigns a review comment's id only on POST and inline comments are append-only, so a comment cannot link to itself. There is nothing to fill it with. The fix is therefore to scope the claim rather than repair the divergence:

> **One flag governs both placements**: there is no per-placement opt-out, so the *flag* can never produce a report offering *Fix all* above findings with no *Fix this*.

…followed by a paragraph naming the one identity state that diverges them, in both owners. Stated explicitly because an unqualified absolute invites the opposite fix — inventing a self-link that cannot exist.

`G32o` guards both halves in both files: the phrase must be flag-scoped, and the divergence must be documented. Two checks rather than one, because the phrase is what a reader quotes and the exception is what makes it true. Probed by restoring the absolute in the agent body (first check fails) and by deleting the divergence paragraph from the rule file (second fails).

You filed this as the third confirmed instance of the undischarged-branch pattern on this file, which is fair — all three were an absolute written before its own exception.

---

agents/shared/rules/agent0-fix-links.md
mthines: Correct, and pointed. `20cf328` existed to make the figure re-derivable and then cited a PR for a number that isn't that PR's — 195 is #168 with a 13-character bot login, and its actual reviewer login gives 189. Re-measured:

| Fill | in-body | over 140 |
| --- | --- | --- |
| #168 + `mthines` | 189 | 49 |
| #168 + `claude[bot]` | 197 | 57 |
| #168 + `dash0-dev[bot]` | 200 | 60 |

Rather than swap one un-reproducible number for another, `50db0d7` ships the recipe:

```bash
node agents/pr-reviewer/scripts/build-agent0-link.mjs --env production --source fix-all \
  "/pr-fix https://github.com/mthines/agent-skills/pull/168 <login>" \
  | perl -pe 's/&/&amp;/g' | tr -d '\n' | wc -c
```

Executed before committing: it reproduces 189, 197, and 200 exactly. The paragraph now states the range across realistic fills (189–204), names the login as the only variable and its +4 `development` offset, and records the miscitation so the next edit doesn't repeat it.

Agreed the argument is untouched: the 164 floor already exceeds 140, so the conclusion never depended on which fill the illustration used — which is precisely why the wrong number survived a proofread.

Independently, the relay transform is now confirmed by observation rather than arithmetic alone. Posting the rendered button through this write path and reading it back showed a code span opening inside the anchor's URL attribute; the same markup with a 44-char URL came back byte-for-byte intact. So the discriminator is URL length, exactly as `RELAY_SAFE_URL_MAX` claims, and the table is in the PR description.

---

agents/shared/rules/agent0-fix-links.md
mthines: Fixed in `50db0d7` — the paragraph was rewritten for the finding above anyway, and now wraps at 102 max, in band with its 95–101 neighbours. Diagnosis was exact: the previous edit re-flowed a sentence boundary and left the line long.

---

agents/pr-reviewer.md
```diff
@@ -2259,10 +2261,24 @@ can `PATCH`. Write each rendered body to a file and run the shared checker on it
...
+# long URLs; the `gh` path sends a file and rewrites nothing, so asking there withholds a button
+# that would have posted intact. Every fix link is over budget by construction, so an
+# unconditional check is a permanent, silent opt-out of the feature on every path
+# (agent0-fix-links.md § Relay length limit). This block is a separate tool call from Step 4a, so
+# resolve the path here with github-access.md § Step 0's own probe, as with $AGENT_MD above.
+# Default to `mcp`: an inconclusive probe must WITHHOLD (a mangled button is worse than none),
+# so the fail-safe direction is "assume relayed", never "assume file".
+ACCESS_PATH=mcp
+command -v gh >/dev/null 2>&1 \
+  && gh api "repos/$RESOLVED_REPO" --jq .full_name >/dev/null 2>&1 \
```
mthines: issue (high): 🟠 **Write-path probe reads $RESOLVED_REPO without binding it** **(blocking)**

Shell state dies between tool calls, so the probe hits `repos/` and 404s: `ACCESS_PATH` stays `mcp` and the buttons are withheld on `gh` too. § Step 0 derives the repo and guards `-n`.

Evidence: `agents/shared/rules/github-access.md:43` (derives + guards) · `agents/pr-reviewer.md:329` (only binding) · `agents/pr-reviewer.md:2703` (same copy in 4a)

```bash
TARGET_REPO="${RESOLVED_REPO:-$(gh repo view --json nameWithOwner -q .nameWithOwner 2>/dev/null)}"
command -v gh >/dev/null 2>&1 && [ -n "$TARGET_REPO" ] \
  && gh api "repos/$TARGET_REPO" --jq .full_name >/dev/null 2>&1 && ACCESS_PATH=gh
```

<sup>`pr-reviewer` · commit `e8c62d0`</sup>



---

agents/pr-reviewer.md
```diff
@@ -2675,22 +2692,36 @@ obligations, and the first is now the load-bearing one:
...
-     [ "$rc" -eq 3 ] && NOTE_MANGLED_LINK=1
+     case $? in
+       1) RERENDER_WITH_NO_FIX_LINKS=1 ;;   # re-render, post that, and say so in the run line
+       3) NOTE_MANGLED_LINK=1 ;;            # not remediable here — post, and note it
+     esac
+
+     # 1 wins over 3 when a body carries BOTH kinds — one remediable URL makes the RUN remediable,
+     # not the body clean — so the exit-3 condition can SURVIVE the withhold. Re-render first, then
+     # ask again, or a mangled citation goes unnamed on exactly the runs that had two problems.
+     if [ -n "$RERENDER_WITH_NO_FIX_LINKS" ]; then
```
mthines, line 2734: issue (medium): 🟡 **Exit-3 re-check runs on the un-re-rendered body** **(non-blocking)**

Only a flag is set here, so the second `--relay-check` reads the same file and returns 1 again — `NOTE_MANGLED_LINK` never gets set. The inline sibling re-renders before re-asking.

Evidence: `agents/pr-reviewer.md:2290` (inline re-renders first) · `agents/pr-reviewer/scripts/comment-spine.mjs:331` (1 wins over 3)

```bash
# measured on a body carrying BOTH an over-budget fix link and an over-budget citation:
#   --relay-check  -> 1
#   --relay-check  -> 1   (same file, nothing re-rendered)
#   --relay-check  -> 3   only once the fix link is actually gone
```

<sup>`pr-reviewer` · commit `e8c62d0`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -2796,105 +2803,175 @@ const isPollBlock = (block) =>
...
+        : "the call is unconditional, so the buttons are withheld on the `gh` path too and the"
+          + " affordance never renders anywhere");
   }
+  s.check("G32p github-access.md binds ACCESS_PATH with how the body travels",
+    (() => {
+      const ga = readFileSync(join(REPO_ROOT, "agents/shared/rules/github-access.md"), "utf8");
+      return /`ACCESS_PATH`/.test(ga) && /tool-call argument/.test(ga) && /Body travels as/.test(ga);
+    })(),
+    "the consumers read $ACCESS_PATH to decide whether their body is relayed — unbound, the"
+      + " comparison is always false and every write is treated as file-based");
```
mthines: suggestion (low): ⚪ **Guard message names the opposite fail-safe direction** **(non-blocking)**

Both consumers initialise `ACCESS_PATH=mcp`, so a failed probe treats every write as relayed and withholds the button — not as file-based. The block's own comment says so four lines up.

Evidence: `agents/pr-reviewer.md:2270` (assume relayed, never file) · `agents/pr-reviewer.md:2272` (ACCESS_PATH=mcp default)

```text
unbound, the comparison is always false and every write is treated as relayed
```

<sup>`pr-reviewer` · commit `e8c62d0`</sup>



---

agents/shared/rules/agent0-fix-links.md
```diff
@@ -404,6 +465,19 @@ on a branch that gets deleted.
 **Some write paths cannot carry these buttons at all, and the correct response is to withhold them
 — never to shorten the prompt until it fits.**
 
+**`--relay-check` is asked only on those paths — the condition is the feature.** Gate it on
+`ACCESS_PATH == "mcp"` ([`github-access.md § Step 0`](./github-access.md#step-0--resolve-your-path-once-before-any-github-step)),
```
mthines: nitpick (low): This line is 130 chars where its own paragraph runs 91–104 and the file's p90 is 101 — break before the long section link. **(non-blocking)**

<sup>`pr-reviewer` · commit `e8c62d0`</sup>

---

mthines (commented):
Reviewed `e8c62d0` — 4 findings — 1 blocking — need attention before human review. 4 finding(s) inline. Sticky exists but this MCP access path has no issue-comment update tool — report not updated in place.

---

agents/pr-reviewer.md
mthines: Confirmed, and this is the sharpest finding of the run: the fix in `e8c62d0` did not work at all. Fixed in `8a20a5d`.

`RESOLVED_REPO` is bound at exactly one site — Step 0.2 — and both of my blocks state in their own comments that they are a separate tool call. So the variable was empty in both, the probe hit `repos/`, 404'd, `ACCESS_PATH` stayed `mcp`, and the buttons were withheld on the `gh` path exactly as before. The gate was the unconditional check wearing an `if`.

Worse than an oversight: `github-access.md § Step 0` carries both protections I dropped — the derivation chain with fallbacks, and a comment naming `RESOLVED_REPO` specifically as the caller variable not to assume is bound, because "an unset one probes `repos/` and 404s, which reports 'no gh' on a session where gh works". I copied the probe out of the rule that predicts this failure and left the prediction behind.

Both sites now derive it:

```bash
TARGET_REPO="${RESOLVED_REPO:-$(gh repo view --json nameWithOwner -q .nameWithOwner 2>/dev/null)}"
ACCESS_PATH=mcp
[ -n "$TARGET_REPO" ] && command -v gh >/dev/null 2>&1 \
  && gh api "repos/$TARGET_REPO" --jq .full_name >/dev/null 2>&1 \
  && ACCESS_PATH=gh
```

The departure from § Step 0's "undecided ⇒ defer the probe" rule is deliberate and now stated in both blocks: the write happens inside them, so there is no later step to defer to, and an undecided probe must still answer — `mcp`, because withholding is recoverable and a mangled button is not.

Your closing point is the one I've acted on hardest. `G32p` asserted the gate existed, which was true the whole time, so it passed while the feature stayed broken — a guard measuring the shape of the fix instead of its effect. It now also asserts the probe can reach `gh` (no bare `repos/$RESOLVED_REPO`, and a derivation present whenever a `repos/` probe is). Probed by restoring the old probe: both site checks fail.

Noting for the record why `G43b` was blind here — it asks whether a name is bound anywhere in the file, and this one is. Per-block binding is a different question, and this is the first case that has needed it.

---

agents/pr-reviewer.md
mthines: Correct — the branch was dead code that read as coverage. Fixed in `8a20a5d`.

The comment said "Re-render first, then ask again" and the shell re-checked `/tmp/report-body.md` unchanged, which returns the same exit 1 forever. So `NOTE_MANGLED_LINK` was unreachable from that path, and a mangled citation went unnamed on precisely the runs that had two problems — the case the branch was written for.

It now does what its comment claims:

```bash
jq 'del(.FIX_ALL_URL)' /tmp/report-payload.json > /tmp/report-payload.nofix.json
node "$RENDER" /tmp/report-payload.nofix.json > /tmp/report-body.md || abort …
node … --relay-check /tmp/report-body.md
```

which is the shape the inline block at Step 2.8 already had — it re-rendered through `del(.FIX_URL)` before re-asking. The two call sites had drifted, and the report side was the wrong one.

Guarded positionally: `del(.FIX_ALL_URL)` must appear in the block *before* the last `--relay-check`. A presence check alone would pass with the re-render sitting after the re-ask, which is the same defect one line further down.

This is the identical prose-vs-shell split as the blocking finding above — a comment describing the correct behaviour immediately above a shell that does not implement it, twice in one block. That pattern is now what both new guards are built to detect rather than restate.

---

scripts/eval/l1.mjs
mthines: Right, and the inversion matters more than the wording. Fixed in `8a20a5d`.

The message said an unbound `ACCESS_PATH` means "every write is treated as file-based" — but the code assigns `ACCESS_PATH=mcp` before the comparison, so unbound is not reachable, and "file-based" is the opposite of the fail-safe two lines above it. A failure message that names the wrong direction sends the next reader to fix the wrong end.

The real risk this check covers is **token drift**: the consumers compare against the literal `"mcp"`, so renaming the token in `github-access.md` makes the comparison always-false, `WRITE_IS_RELAYED` stays unset, and every write is treated as file-based — a posted mangled button, which is exactly the direction the `mcp` default exists to avoid. That is now what the message says.

---

agents/shared/rules/agent0-fix-links.md
mthines: Fixed in `8a20a5d` — the link moved to its own line, so the paragraph is back inside its wrap band.

Fair to flag that this is the second formatting recurrence in this file one commit after `50db0d7` fixed the previous one. Both were the same mechanism: I edited mid-paragraph and left the line I touched unwrapped. Checking the whole edited paragraph rather than the changed sentence is the habit that would have caught either.

---

PR #176, opened 2026-09-05
Convert the aw dispatcher from an agent to a skill

mthines:
`aw` was an agent, and it failed this repo's own third-agent test. [`anthropic-architecture-research.md` §2.2](https://github.com/mthines/agent-skills/blob/main/skills/workflow/autonomous-workflow/references/anthropic-architecture-research.md) says a new agent must demonstrate a third independent *context*, not a third *role* — and the dispatcher's boundary is conditional: real when routing.rule.md auto-triggers it mid-conversation, absent when a developer opens a session specifically to type `/aw`. The planner/executor boundary is unconditional; the dispatcher's was not.

Being an agent also cost it a delegation rung. With `aw` at rung 1, `aw-planner` and `aw-executor` sat at rung 2 and held no `Task` tool, which made four documented behaviours unreachable: the Phase 1 `Explore` fan-out, the Phase 3 file-disjoint fan-out, the Phase 7 `ci-auto-fix` fan-out, and `create-pr`'s post-draft `review-loop` pass. That last one is not theoretical — the `aw-lessons` bucket carries a field report of `NOT REVIEWED` draft PRs shipping because of it.

**What changed.** `templates/aw.agent.md` became `aw/SKILL.md` (a nested skill, installed as `~/.claude/skills/aw`). It declares no `tools:` and no `model:`: it inherits the caller's grant and runs in the caller's context, so the two agents it dispatches now sit at the top-level rung and keep their own `Task`. `routing.rule.md` dispatches `Skill("aw", …)` instead of `Task(subagent_type="aw", …)`, and its dispatch-unavailable branch is gone — a skill needs no `Task` to start.

The tier table is now single-sourced in `SKILL.md` Step 1. The old guard byte-compared two copies; `G2b` now asserts the dispatcher carries *no* tier rows and links the canonical table instead. New `G24b` asserts the dispatcher documents tool-grant inheritance and degradation — the thing an agent's `tools:` block used to make structural.

**The install path was the sharp edge, and it was sharper than first diagnosed.** `install.sh` removes the pre-v3.23 `aw` *agent* symlink in every mode, but the first draft created the replacement *skill* link only under `--development` — so a `--global`/`--project` upgrade deleted `/aw` and created nothing. Hoisting that link out of the dev block exposed the general case: **skills are discovered under a flat installed name, so no nested `SKILL.md` is reachable by the normal skill-install path** — and `aw` was not the only one. `aw-setup` and `aw-tester-chrome` were never linked either. `aw-setup` is the one that hurts: `install.sh`'s own summary tells the user to run `/aw-setup`, `aw-planner` **halts** on a UI task until an aw-target exists, and `phase-1-planning.md` forbids the planner from scaffolding one itself — three surfaces pointing at a skill that was never installed. The existence guard, the dev-mode cross-tool hop, and the Claude-side link are now one `NESTED_SKILLS` loop covering all three, and `sync-symlinks.sh`'s `--aw` bundle (which omitted the same two) matches.

**One class of error recurred four times, so it is now guarded.** Moving `aw` out of the agent set made "how many `aw-` agents are there" a question the docs kept answering differently — and three successive fixes each corrected the flagged line and introduced a fresh instance elsewhere. New **`G2c`** derives the truth from `install.sh`'s own `ln -sf … agents/aw-*.md` calls, cross-checks that extraction against `templates/aw-*.agent.md` on disk (so an under-read fails loudly instead of making every downstream assertion pass against a wrong number), then asserts every prose count claim equals it across 9 sites in 4 files, that every installed agent is inventoried in root `CLAUDE.md`, and — pinned to the live count, not a loose floor — that those claim sites still exist.

**Docs beyond the count claims.** Four surfaces outside the changed set still described the pre-v3.23 shape and are not reachable by `G2c`: `scripts/eval/README.md` and `eval-iterate/rules/eval-bug-classification.md` both named the `tier-routing` rubric source as the dispatcher's own tier table (`l2.mjs` reads `SKILL.md § Step 1`, and the dispatcher deliberately has none); §2.2 of the research doc states the third-agent test without recording that it was applied, which this PR is the one case of; and §3.2 claimed "the agent count to two" while the installer links three — now scoped to the *pipeline*, which is what that anti-pattern argument is actually about.

**`aw` is now the only natural-language entry point (v3.24).** Making the dispatcher a skill left one thing unfinished: the parent `autonomous-workflow` skill's description still advertised the same trigger vocabulary — `"autonomously"`, `"independently"`, `"end-to-end"`, `"in isolation"`, `"take care of this"` — so two skills competed for one request with nothing to arbitrate. Losing that race is not cosmetic: a request landing on the parent runs the phase machinery with **tier detection skipped** (a Micro typo fix pays the Full planner→executor handoff, or a Full task runs single-pass with no `plan.md` and no `confidence(plan)` gate) and the **universal lessons loop skipped** — the two things the dispatcher exists to own, and exactly what v3.12 hoisted to it. The parent's description now describes the phase machinery, states it is not the entry point, and names `aw`, while quoting no trigger phrase and making no `Triggers on` / `Use when` claim; `/autonomous-workflow` still runs the machinery directly for a deliberate bypass, and the inventory marker moves `auto` → `/` to say so.

The routing rule keeps its **full** positive trigger list rather than becoming a pointer. The duplication is forced by two consumers: frontmatter cannot link a canonical home the way `aw/SKILL.md` links the tier table, and that file is *also* the live rubric the L2 `aw-should-trigger` suite reads — slimming it would leave that suite with no vocabulary to decide `trigger` on. So it is kept honest by assertion instead. New **`G2d`** guards both halves: the parent quotes no trigger phrase and makes no trigger claim (matched on the *claiming* forms, since the sentence that fixes the problem must name the words it disclaims), and every phrase `aw` advertises exists in the routing rule. The second half fired on its first run — `aw` advertised `"implement autonomously"`, a phrase the rule had never listed contiguously and which was the phrasing just deleted from the parent. Its phrase-count sentinel is **pinned to the live count (14), not a floor** — the loose-floor defect the reviewer caught in `G2c`, which the first draft of `G2d` reproduced despite a comment claiming otherwise; found by mutation, fixed, re-verified.

**One judgment call beyond the shape change:** `aw-planner` and `aw-executor` gained `- Task` in their `tools:` frontmatter. That is what actually reconnects the four fan-outs; it is revertible on its own if you'd rather land the shape change alone.

**Verification.** L1 green at 1342/1342, L2 green in CI. Every new guard was mutation-tested — broken, confirmed to fire, restored. `G2c`'s own first three versions were each defeated by the reviewer with an executed receipt (a hyphenated agent name, a single reworded site against a loose sentinel, a claim site outside the file list); `G2d`'s four assertions were mutation-tested the same way, and its sentinel is the one that did not fire until pinned. `install.sh` was run in a sandboxed `HOME` across **all three modes** with a legacy `aw.md` symlink pre-seeded: legacy agent removed, all three nested skills resolve, all three agents link, re-runs exit 0, `--project` writes nothing into `HOME`, and an incomplete skill directory exits non-zero naming the missing path.

**Known, recorded, not fixed here:** the dispatcher is 294 body lines against its own ~200-line ceiling (it was ~274 as an agent). Recorded in `autonomous-workflow/CLAUDE.md` with the likely fix named — move the single-context Full procedure into a rule file. And the nested-skill linking has no L1 guard: the assertion worth writing ("every nested `SKILL.md` is linked") would derive from the same loop in the same commit, where `G2c` earned its place by deriving from two independent sources. The convention is recorded on-thread instead — a new nested skill goes in `NESTED_SKILLS` *and* `AW_SKILLS`.

https://claude.ai/code/session_01WJcLUmKMNeYXkG6xzosAtR


mthines:
### ✅ No issues found

Small delta (69 lines) that fixes the third aw- agent-count drift with a real edit and adds a mechanical L1 guard (G2c) tying every prose count claim to install.sh.

<details>

| Gate | Status | Details |
|---|---|---|
| Description vs. code | ✅ | Description reflects the PR's overall intent (aw agent→skill conversion). |
| Prior review feedback | ✅ | 9 resolved since `5980319`; 0 open. |
| Documentation | ✅ | Commit messages and inline replies document the fix. |
| Self-review signals | ✅ | No debug/TODO/FIXME markers in the delta. |
| Code review | ✅ | No findings in this delta; L1 1336/1336 executed against PR head. |

**Found**

Quality — produced 0 candidates on the delta → posted inline 0.
Verified — node scripts/eval/l1.mjs executed against the PR head: 1336/1336; the new G2c regex verified against the live install.sh (returns exactly [aw-planner, aw-executor, aw-tester]).

**Run**

incremental · 69 lines in delta · tier standard · depth checkout · cumulative 150 lines since the last full-tier pass (FULL_REFRESH_DELTA=150, not exceeded)
CI — green (l1, l2 both pass)
Memories — 6 indexed · 2 used (2 hotspots)

- **hotspot** `hotspot::scripts/eval/l1.mjs` — test-gap hotspot recorded this class recurring 3x; this run confirms the G2c guard closes it
- **hotspot** `hotspot::README.md` — missing-update hotspot (3 prior instances); this run confirms all count-claim sites now agree at three

<sup>Nothing to report — standards (2 docs), optimality (skipped), measurability (skipped), integrations (not activated), severity, 0 files skipped.</sup>

</details>

<sup>`pr-reviewer` · commit `f55eca5` · incremental review, delta since `5980319` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md) · updated <relative-time datetime="2026-09-05T06:55:10.000Z">Sep 5, 2026 6:55am UTC</relative-time></sup>

skills/workflow/autonomous-workflow/install.sh
```diff
@@ -182,13 +185,35 @@ if [[ "$MODE" == "development" ]]; then
...
+#   autonomous-{planner,executor}.md — pre-`aw-` namespace names.
+#   aw.md                            — the dispatcher was an agent until v3.23;
+#                                      it is now the `aw` SKILL linked above.
+#                                      Leaving the agent in place would put two
+#                                      `aw` entries with near-identical
+#                                      auto-trigger descriptions in the harness.
 for legacy in "autonomous-planner.md:planner.template.md" \
-              "autonomous-executor.md:executor.template.md"; do
+              "autonomous-executor.md:executor.template.md" \
+              "aw.md:aw.agent.md"; do
```
mthines, line 262: issue (high): 🟠 **aw skill link is dev-only but the agent removal is not** **(blocking)**

This removal runs in every mode, yet the replacement `aw` skill is linked only inside the `--development` block and is missing from both `npx skills add` lists this script prints.

Evidence: `skills/workflow/autonomous-workflow/install.sh:200` (link is dev-only) · `skills/workflow/autonomous-workflow/install.sh:126` (recovery list omits aw)

```sh
# install.sh:127 and :136 — give a --global/--project user a way back to /aw
echo "    --skill autonomous-workflow aw aw-create-plan aw-create-walkthrough confidence \\" >&2
```

<sup>`pr-reviewer` · commit `1a3d640`</sup>



---

skills/workflow/autonomous-workflow/README.md
```diff
@@ -135,22 +135,27 @@ drop the `--aw` flag. To preview without applying, add `--dry-run` (or `-n`).
...
-Then say *"implement X independently"* (or invoke `@aw`) — the routing rule
-dispatches the **`aw` dispatcher**, which detects the tier and routes: Micro/Lite
+Then say *"implement X independently"* (or invoke `/aw`) — the routing rule
+invokes the **`aw` dispatcher**, which detects the tier and routes: Micro/Lite
 run single-pass; Full hands off to the planner→executor split.
 
 #### What gets installed
 
-Four agents linked into your `.claude/agents/` directory under the
+The **`aw` dispatcher** as a skill in your `.claude/skills/` directory — it runs
```
mthines, line 144: issue (medium): 🟡 **New install block contradicts the summary above it** **(non-blocking)**

Line 128 still lists `aw` among the agents `--aw` links, which this block correctly reclassifies as a skill. The `SKILL.md` twin of this claim was fixed; the README one was not.

Evidence: `skills/workflow/autonomous-workflow/README.md:128` (stale agent claim)

```md
`--aw` symlinks the autonomous-workflow skill, the `aw` dispatcher skill, and its
companion skills, plus the `aw-planner` / `aw-executor` agents, into your
`~/.claude/` directory.
```

<sup>`pr-reviewer` · commit `1a3d640`</sup>



---

skills/workflow/autonomous-workflow/templates/aw-planner.agent.md
```diff
@@ -77,7 +83,7 @@ finished — even if the work feels done.
    `SKILL.md` Auto-Trigger Setup section). Do not attempt to plan without it.
 
 2. **Detect workflow mode** — output the canonical block (field-for-field
-   identical to `SKILL.md` Step 1 and the dispatcher template) before doing
+   identical to `SKILL.md` Step 1, which is the tier table's only home) before doing
```
mthines, line 86: issue (low): ⚪ **`SKILL.md` Step 1 keeps the twin dispatcher-template claim** **(non-blocking)**

The same sentence was corrected here but not at `SKILL.md:114`, which still credits "the dispatcher template" for emitting the block. That section is now the live L2 `tier-routing` rubric.

Evidence: `skills/workflow/autonomous-workflow/SKILL.md:114` (stale twin) · `scripts/eval/l2.mjs:26` (reads that section)

<sup>`pr-reviewer` · commit `1a3d640`</sup>



---

mthines (commented):
---

skills/workflow/autonomous-workflow/install.sh
mthines: Fixed in `8cfabd3`, but **not with the suggested patch** — the finding is right, the proposed fix isn't.

Adding `aw` to the two `npx skills add --skill` lists assumes that CLI resolves a *nested* skill. I couldn't establish that: `aw-setup` and `aw-tester-chrome` are the two sibling nested skills in this repo and neither appears in any `npx skills add` list anywhere. And root `README.md:255` documents the `--global` path as `npx skills add` **followed by** `install.sh --global`, so the frozen copy already ships `aw/` — what it can't do is surface it as `/aw`, because skills are discovered under a flat installed name.

So the repair is in this script rather than in a hint about another tool: the Claude-side link and its `mkdir -p` are hoisted out of the `--development` block and now run in every mode. The cross-tool `~/.agents/skills/aw` hop stays development-only (it points at the clone — wrong tier for a project-local install), and the link falls back to `$SKILL_DIR/aw` when that hop wasn't created.

Two follow-ons in the same sweep:
- `cc9a8c5` adds a sanity check for `aw/SKILL.md`. Making the link unconditional over a frozen copy meant an incomplete install would produce a *dangling* `/aw` symlink instead of an error.
- Root `README.md:255` was a third stale twin of the finding below it ("links the three `aw-` agents") and is corrected in `8cfabd3`.

Verified in a sandboxed `HOME` across all three modes with a legacy `aw.md` symlink pre-seeded: legacy agent removed, `.claude/skills/aw/SKILL.md` resolves, re-run idempotent. The new guard was tested by deleting `aw/` from a copied skill directory — it exits non-zero naming the path and creates no link.

---

skills/workflow/autonomous-workflow/README.md
mthines: Fixed in `dd0f257`.

The companion count had to move with it: `AW_SKILLS` in `sync-symlinks.sh` holds 14 entries — `autonomous-workflow`, `aw`, and 12 companions — so `aw` crossing from the agents half of the sentence to the skills half leaves "12 companion skills" correct rather than off by one.

The sweep also turned up a third instance of the same stale claim at root `README.md:255` ("links the three `aw-` agents"), fixed in `8cfabd3`.

---

skills/workflow/autonomous-workflow/templates/aw-planner.agent.md
mthines: Fixed in `7f6c83d` — the block now credits the `aw` dispatcher skill instead of the retired template.

Confirmed the second-order point: `extractSection` does pull this section for the L2 `tier-routing` suite (the rubric was repointed here in `1a3d640`), so the stale sentence was being fed to the model on every run of that suite, not just sitting in prose.

---

README.md
```diff
@@ -251,7 +252,7 @@ Full-tier plans are **grounded, traceable, and executable** (v3.15): every plann
 
 ### Setup
 
-The [clone + symlink](#recommended-clone--symlink) install already links the three `aw-` agents and the routing rule — `sync-symlinks.sh` runs the dispatcher installer for you. If you used `npx skills add` instead, run the dispatcher installer manually:
+The [clone + symlink](#recommended-clone--symlink) install already links the `aw` dispatcher skill, the two `aw-` agents, and the routing rule — `sync-symlinks.sh` runs the dispatcher installer for you. If you used `npx skills add` instead, run the dispatcher installer manually (it places the `aw` dispatcher, which a plain skill install cannot reach — it is nested inside the autonomous-workflow skill):
```
mthines: issue (medium): 🟡 **`aw-` agent count drops to two; three are linked** **(non-blocking)**

`install.sh` links three `aw-` agents — planner, executor and tester — which I confirmed by running all three install modes. Line 328 of this same file still says three.

Evidence: `README.md:328` (still says three) · `skills/workflow/autonomous-workflow/install.sh:262` (aw-tester link)

```md
install already links the `aw` dispatcher skill, the three `aw-` agents, and the routing rule
```

<sup>`pr-reviewer` · commit `de92af4`</sup>



---

README.md
```diff
@@ -174,11 +174,12 @@ Meta — scaffolding new skills, maintaining docs, persisting memory.
...
 
-The flagship `aw` agents are **generated from templates** in `skills/workflow/autonomous-workflow/templates/` (each template's filename matches its installed agent name) and symlinked into `~/.claude/agents/` by `install.sh` — they are not stored as `agents/*.md`, so search the templates directory to find them. See [Featured: autonomous workflow](#featured-autonomous-workflow) for the full picture.
+The flagship `aw-*` agents are **generated from templates** in `skills/workflow/autonomous-workflow/templates/` (each template's filename matches its installed agent name) and symlinked into `~/.claude/agents/` by `install.sh` — they are not stored as `agents/*.md`, so search the templates directory to find them. See [Featured: autonomous workflow](#featured-autonomous-workflow) for the full picture.
+
+**`aw` itself is not in this table — it is a skill**, [`/aw`](./skills/workflow/autonomous-workflow/aw/SKILL.md), so it runs in your context rather than in one of its own. That is what lets the agents it dispatches sit at the top-level rung and keep their own sub-agent dispatch.
```
mthines, line 179: issue (medium): 🟡 **`aw` still tabled as an agent further down** **(non-blocking)**

This new line says `aw` is not in the agents table because it is a skill. The `Three agents, one workflow` table at line 230 still lists `aw` under an `Agent` column with an exit gate.

Evidence: `README.md:224` (heading) · `README.md:230` (aw row)

<sup>`pr-reviewer` · commit `de92af4`</sup>



---

skills/workflow/autonomous-workflow/README.md
```diff
@@ -124,9 +124,9 @@ cd agent-skills
 bash scripts/sync-symlinks.sh --aw
 ```
 
-`--aw` symlinks the autonomous-workflow skill and its 12 companion skills,
-plus the `aw` / `aw-planner` / `aw-executor` agents, into your `~/.claude/`
-directory. Edits to the cloned repo are picked up live on the next agent
+`--aw` symlinks the autonomous-workflow skill, the `aw` dispatcher skill, and
+its 12 companion skills, plus the `aw-planner` / `aw-executor` agents, into
```
mthines: suggestion (low): ⚪ **`--aw` agent list omits `aw-tester`** **(non-blocking)**

This rewritten list names two agents while line 149 of the same file says three are linked. Dropping `aw` was right; `aw-tester` should have taken its place.

Evidence: `skills/workflow/autonomous-workflow/README.md:149` (says three)

```md
its 12 companion skills, plus the `aw-planner` / `aw-executor` / `aw-tester`
```

<sup>`pr-reviewer` · commit `de92af4`</sup>



---

mthines (commented):
Reviewed `de92af4` — 3 non-blocking findings; the substituted install fix itself verifies correct in all three modes. 3 finding(s) inline. Sticky exists but this access path (GitHub MCP, no gh CLI) has no issue-comment update tool — report not updated in place.

---

README.md
mthines: Fixed in `5980319` — reverted to "three".

You're right, and this one was a regression my own sweep introduced rather than a pre-existing gap: `aw` was never `aw-` prefixed, so "three `aw-` agents" already meant planner/executor/tester and needed no edit at all. Confirmed against the `ln -sf` calls at `install.sh:256,259,262`.

---

README.md
mthines: Fixed in `5980319`.

Retitled to "One dispatcher, two agents", gave the table a `Kind` column so `aw` reads as `skill` rather than sitting under `Agent`, and rewrote the trailing sentence — it claimed all three share the `aw-` prefix and group in `~/.claude/agents/`, which is now false for `aw` on both counts. Checked for inbound links to `#three-agents-one-workflow` before renaming; there are none.

---

skills/workflow/autonomous-workflow/README.md
mthines: Fixed in `5980319` — `aw-tester` takes the slot `aw` vacated.

The same block had a fourth instance you didn't flag, fixed in the same commit: ":149" says "Plus three agents" and then introduces a table whose **first row is the `aw` skill** — four rows, not three. The table is now introduced as all four units, so the sentence and the thing it introduces agree.

---

README.md
```diff
@@ -216,21 +217,21 @@ The rest of this section is the under-the-hood reference.
...
+| `aw-planner`  | agent | Full tier, 0–2 (validate, plan, worktree + `plan.md`) | `confidence(plan) ≥ 90%` |
+| `aw-executor` | agent | Full tier, 3–7 (implement, test, docs, PR, CI) | CI green, walkthrough delivered |
 
-All share the **`aw-`** prefix ("autonomous-workflow"): deliberate namespace so they group together in `~/.claude/agents/` and disambiguate from agents installed by other skills. `aw` is adaptive — it only invokes the planner→executor split for **Full** tasks; Micro/Lite run single-pass.
+The two agents share the **`aw-`** prefix ("autonomous-workflow"): deliberate namespace so they group together in `~/.claude/agents/` and disambiguate from agents installed by other skills. `aw` carries no prefix and does not live there at all — it is a skill in `~/.claude/skills/`, which is precisely what lets the agents it dispatches sit at the top-level rung and keep their own sub-agent dispatch. `aw` is adaptive — it only invokes the planner→executor split for **Full** tasks; Micro/Lite run single-pass.
```
mthines: issue (medium): 🟡 **`aw-` prefix set is three agents, not two** **(non-blocking)**

Three `aw-` agents are linked into `~/.claude/agents/` — planner, executor and tester. Lines 208, 255 and 328 of this same file say three; this line and the heading at 224 say two.

Evidence: `skills/workflow/autonomous-workflow/install.sh:262` (aw-tester ln -sf into agents/) · `README.md:255` (same file says three)

```md
### One dispatcher, three agents                       <- line 224
The three agents share the **`aw-`** prefix ...        <- line 234
```

<sup>`pr-reviewer` · commit `5980319`</sup>



---

CLAUDE.md
```diff
@@ -98,10 +98,12 @@ Type markers (by primary entry point — all three are technically model-invocab
...
+
+The agents it hands off to are **generated from templates**, not stored as `agents/*.md`, so searching `agents/` for them returns nothing — search `skills/workflow/autonomous-workflow/templates/` instead (each template's filename matches its installed agent name):
 
-- `aw` — opt-in dispatcher: reads `aw-lessons`, detects tier (Micro/Lite/Full), routes single-pass vs the planner→executor split. Source: [`templates/aw.agent.md`](./skills/workflow/autonomous-workflow/templates/aw.agent.md), installed by `install.sh` as `~/.claude/agents/aw.md`
 - `aw-planner` — Full tier, phases 0–2 (validate, plan, worktree + `plan.md`), gated on `confidence(plan) ≥ 90%`. Source: [`templates/aw-planner.agent.md`](./skills/workflow/autonomous-workflow/templates/aw-planner.agent.md), installed as `aw-planner.md`
 - `aw-executor` — Full tier, phases 3–7 (implement, test, docs, PR, CI). Source: [`templates/aw-executor.agent.md`](./skills/workflow/autonomous-workflow/templates/aw-executor.agent.md), installed as `aw-executor.md`
```
mthines, line 108: question (low): `aw-tester.agent.md` sits in this same `templates/` dir and installs as an agent, but is not listed. Deliberate scoping to aw's own handoffs, or the same omission the READMEs had? **(non-blocking)**

<sup>`pr-reviewer` · commit `5980319`</sup>

---

scripts/eval/l1.mjs
```diff
@@ -2609,7 +2625,28 @@ const isPollBlock = (block) =>
...
+
+  // G24b — the skill-shaped counterpart of G24. `aw` reaches GitHub through
+  // create-pr / review-loop but has no `tools:` block to grant anything, so its
+  // contract is that a missing grant is named in the terminal report, never
+  // silently swallowed.
+  // Collapse whitespace first: this repo uses semantic line breaks, so a phrase
+  // that reads as one sentence is routinely split across lines mid-clause.
+  const awSkill = readFileSync(join(AW, "aw/SKILL.md"), "utf8").replace(/\s+/g, " ");
+  s.check(
+    "G24b aw dispatcher skill documents tool-grant degradation",
```
mthines, line 2804: suggestion (medium): 🟡 **L1 has no guard tying the aw agent count to its installer** **(non-blocking)**

This class has recurred four times in this PR, and L1 is green at `5980319` while `README.md` line 234 still contradicts 208, 255 and 328. Deriving the count from the `ln -sf` targets would close it.

Evidence: `skills/workflow/autonomous-workflow/install.sh:262` (the ln -sf to count) · `README.md:234` (live contradiction, L1 green)

```js
const linked = (readFileSync(join(AW, "install.sh"), "utf8")
  .match(/ln -sf .*\/agents\/aw-[a-z]+\.md/g) || []).length;
s.check("aw- agent count in prose matches install.sh", prosePasses(linked));
```

<sup>`pr-reviewer` · commit `5980319`</sup>



---

mthines (commented):
Reviewed `5980319` — 🟡 3 findings — the reconciliation landed everywhere except the root README's own `aw-` namespace sentence, which still says two where three are linked. 3 finding(s) inline. This access path (GitHub MCP, no `gh`) cannot edit an issue comment, so the sticky report at #issuecomment-5549831638 is NOT updated for this run — the full report is in the run's terminal output.

---

README.md
mthines: Fixed in `f55eca5` — and you're right that it was the same subtraction a third time, four seconds after I'd written down why it was wrong.

Line 234 now states the table's scope explicitly (what `aw` itself routes to) and names `aw-tester` as the third prefixed agent, so the scoped table and the unscoped prefix claim no longer read as one number stated twice.

The heading at 224 went too. Your read was that it's scope-defensible, and it is — but a count-bearing heading ten lines above "all three" is the same trap, so it no longer carries a count at all. It was caught by the new guard on its first run, before I'd noticed it myself.

---

CLAUDE.md
mthines: The same omission, not deliberate scoping — added in `f55eca5`.

The list's stated purpose is that searching `agents/` returns nothing and you should search `templates/` instead. `aw-tester` is exactly a file that behaves that way, so leaving it out meant the doc named the problem and then withheld the answer for one of the three cases.

`G2c` now derives the installed set from install.sh and fails if any member is missing from this list, so it can't silently fall out again.

---

scripts/eval/l1.mjs
mthines: Implemented as `G2c` in `f55eca5`. This was the right finding of the three — the other two were instances, this is the class.

It follows your sketch but holds three things rather than one, because a bare count comparison would have passed vacuously the moment someone reworded a sentence:

1. **Prose counts equal the installer.** Three phrasings, each anchored on something that pins it to the `aw-` set (`` `aw-` agents``, `agents linked|installed`, `dispatcher[ skill], N agents`) so unrelated "agents" prose in the same files isn't swept in. Currently 8 claim sites across `README.md`, `autonomous-workflow/README.md` and `install.sh`.
2. **Every installed agent is inventoried in root `CLAUDE.md`** — otherwise `agents/` stays the wrong place to look and the doc offers no right one.
3. **A sentinel on the claim count** (`>= 5`), so drifting phrasings fail loudly instead of leaving assertion 1 with nothing to check.

All four assertions mutation-tested: prose drift → fires; removing the `aw-tester` `ln -sf` → fires on all 8 sites at once; dropping the CLAUDE.md entry → fires; rewording every phrasing away → the sentinel fires at `found 1`.

It earned itself immediately — its first run failed on the `README.md:224` heading, which neither of us had flagged.

One deliberate departure from the sketch: `linked` is a `Set` of agent *names* rather than a match count, because assertion 2 needs the identities and a duplicate `ln -sf` would otherwise inflate the number.

---

scripts/eval/l1.mjs
```diff
@@ -80,19 +80,99 @@ const REPORT_FIXTURES = (() => {
...
+// two review rounds — each fix corrected the flagged line and left a twin, in a
+// different file, saying the other number. The count is mechanically derivable
+// from the installer, so derive it and make every prose claim answer to it.
+//
+// `aw` is deliberately NOT in the population: it is a skill, has no `aw-`
+// prefix, and is linked into skills/ rather than agents/. That distinction is
+// exactly what the prose kept losing.
+{
+  const installer = readFileSync(join(AW, "install.sh"), "utf8");
+  const linked = [...installer.matchAll(/ln -sf[n]?\s+\S+\s+"\$CLAUDE_DIR\/agents\/(aw-[a-z]+)\.md"/g)]
```
mthines: issue (medium): 🟡 **Truth regex is blind to hyphenated agent names** **(non-blocking)**

`aw-[a-z]+` misses `aw-ui-tester` or `aw-tester2`, so a fourth linked agent stays out of `truth` and all four assertions pass green while every doc undercounts.

Evidence: `scripts/eval/l1.mjs:126` (class excludes hyphen and digits) · `skills/workflow/autonomous-workflow/install.sh:262` (mutation left L1 green)

```js
const linked = [...installer.matchAll(
  /ln -sf[n]?\s+\S+\s+"\$CLAUDE_DIR\/agents\/(aw-[\w-]+)\.md"/g)].map((m) => m[1]);

// Cross-check the extraction itself, so a link this regex cannot see still fails:
const onDisk = readdirSync(join(AW, "templates"))
  .filter((f) => /^aw-.+\.agent\.md$/.test(f)).map((f) => f.replace(/\.agent\.md$/, ""));
s.check("G2c extracted set == templates/ on disk",
  onDisk.length === truth.size && onDisk.every((a) => truth.has(a)),
  `installer: ${[...truth]} | disk: ${onDisk}`);
```

<sup>`pr-reviewer` · commit `f55eca5`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -80,19 +80,99 @@ const REPORT_FIXTURES = (() => {
...
+        claimsSeen++;
+        if (n !== truth.size) wrong.push(`${label}: "${m[0].trim()}" (linked: ${truth.size})`);
+      }
+    }
+  }
+  s.check("G2c prose aw- agent counts match the installer", wrong.length === 0,
+    wrong.length ? wrong.join(" | ") : `${claimsSeen} count claims agree on ${truth.size}`);
+
+  // Sentinel: if the phrasings drift, the check above passes vacuously.
+  s.check("G2c found the aw- agent count claims to guard", claimsSeen >= 5, `found ${claimsSeen}`);
```
mthines: issue (medium): 🟡 **Sentinel bar of 5 leaves three claim sites of slack** **(non-blocking)**

Eight sites match today, so three can drift out silently. Rewording one to "the two specialist agents" both un-matches and mis-states the count, and L1 still passes.

Evidence: `scripts/eval/l1.mjs:175` (eight claims against bar five) · `README.md:255` (reworded alone, L1 green)

```js
s.check("G2c found the aw- agent count claims to guard",
  claimsSeen >= 8, `found ${claimsSeen}`);
```

<sup>`pr-reviewer` · commit `f55eca5`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -80,19 +80,99 @@ const REPORT_FIXTURES = (() => {
...
+  const toNum = (w) => (/^\d+$/.test(w) ? Number(w) : WORDS[w.toLowerCase()]);
+  const CLAIMS = [
+    // "the three `aw-` agents", "links three `aw-` agents"
+    /(\w+)\s+`aw-`\s+agents/g,
+    // "Plus three agents linked into", "three agents installed"
+    /(\w+)\s+agents\s+(?:linked|installed)/g,
+    // "a dispatcher, three agents, eight phases"
+    /dispatcher(?:\s+skill)?,\s+(\w+)\s+agents/g,
+  ];
+  const SURFACES = [
```
mthines, line 180: suggestion (low): ⚪ **A fourth live claim site sits outside SURFACES** **(non-blocking)**

`autonomous-workflow/SKILL.md:294` says "one dispatcher skill and three agents". It sits in neither this list nor any CLAIMS phrasing, so it drifts unguarded.

Evidence: `skills/workflow/autonomous-workflow/SKILL.md:294` (mutation left L1 green)

```js
["autonomous-workflow/SKILL.md", readFileSync(join(AW, "SKILL.md"), "utf8")],

// ...and a CLAIMS phrasing that reaches the "and **three agents**" form:
/dispatcher\s+skill\*{0,2}\s+and\s+\*{0,2}(\w+)\s+agents/g,
```

<sup>`pr-reviewer` · commit `f55eca5`</sup>



---

mthines (commented):
Reviewed `f55eca5` — 3 findings — none blocking; every live `aw-` agent count now matches the installer. 3 finding(s) inline. Sticky exists but this access path (MCP, no `gh`) has no issue-comment update tool — report not updated in place; full body is in the run's terminal output.

---

README.md
mthines: Superseded — see the `G2c` thread. Widened to `aw-[a-z0-9-]+` in `8766952`, and reproduced your receipt: with `aw-ui-tester` linked and templated, the old pattern kept L1 green at 1336/1336 while the installer linked four and all nine prose sites said three. It now fires on every site plus the inventory assertion.

Your all-alpha control (`aw-auditor`) is what made this a blind spot rather than a broken guard, and that distinction is why I also added the second derivation below.

---

scripts/eval/l1.mjs
mthines: Both halves applied in `8766952`. (A reply meant for this thread landed on the resolved `README.md:234` thread by mistake — same content, wrong address.)

Widened to `aw-[a-z0-9-]+`, and added the cross-check, which is the more important half: every assertion trusted one regex over one file, so an under-read made all of them pass against a wrong number — silently, and in the direction that looks green. The disk set is the other end of the same fact, so disagreement now means the extraction is wrong rather than the prose.

Reproduced your receipt first: with `aw-ui-tester` linked and templated, the old pattern kept L1 green at 1336/1336 while the installer linked four and every prose site said three. After the fix it fires on all nine sites plus the inventory assertion.

I also tested the failure mode the cross-check exists for, which the regex widening alone does not cover — rewriting one `ln -sf` target as `"${CLAUDE_DIR}/agents/aw-tester.md"`. That is invisible to any `$CLAUDE_DIR` pattern, and the cross-check catches it: `templated-not-linked: [aw-tester]`.

---

scripts/eval/l1.mjs
mthines: Pinned in `8766952` — to 9 rather than 8, since adding the `SKILL.md` site from the thread below brought the live count up.

The diagnosis is the part worth recording: my own mutation test rewrote *every* phrasing at once and watched the sentinel drop to `found 1`, which is exactly why three sites of headroom went unnoticed. One site at a time is how this class actually behaves — every instance in this PR was a single line — so an extreme-case mutation was the wrong test for a loose floor. Re-tested with your single-site version: `found 8`, fires.

A floor with headroom is the vacuous case in miniature, which was the thing the sentinel existed to prevent. Asymmetry is deliberate: adding a claim site passes, losing one does not.

---

scripts/eval/l1.mjs
mthines: Both applied verbatim in `8766952` — the surface entry and the fourth pattern.

The doubly-unguarded point was the useful part: adding the file alone would not have reached it, because `**one dispatcher skill** and **three agents**` matches none of the original three patterns. A finding that had named only the missing surface would have produced a fix that still passed the mutation.

Also noting your observation that this repo's own hotspot record already recommended `SKILL.md` as a surface and the implementation substituted `install.sh` for it. That is a fair hit — I read the memory and then didn't use it.

Re-tested with your mutation (`**two agents**` at `SKILL.md:294`): fires, naming the site.

---

skills/workflow/autonomous-workflow/install.sh
```diff
@@ -182,13 +200,47 @@ if [[ "$MODE" == "development" ]]; then
...
+# skill, and this script has to place it. That is load-bearing: the legacy
+# cleanup below removes the pre-v3.23 `aw` AGENT unconditionally, so a
+# --global / --project upgrade that skipped this link would delete /aw and
+# leave no way back.
+mkdir -p "$CLAUDE_DIR/skills"
+if [[ -e "$CLAUDE_DIR/skills/aw" && ! -L "$CLAUDE_DIR/skills/aw" ]]; then
+  echo "error: $CLAUDE_DIR/skills/aw already exists and is not a symlink" >&2
+  exit 1
 fi
+ln -sfn "${AW_LINK_TARGET:-$SKILL_DIR/aw}" "$CLAUDE_DIR/skills/aw"
```
mthines: suggestion (medium): 🟡 **Sibling nested skills stay unreachable outside --development** **(non-blocking)**

This rationale applies verbatim to aw-setup and aw-tester-chrome, which the block does not link. A sandboxed --global install leaves /aw-setup unresolvable while line 292 tells the user to run it.

Evidence: `skills/workflow/autonomous-workflow/install.sh:292` (instructs running /aw-setup) · `skills/workflow/autonomous-workflow/templates/aw-planner.agent.md:196` (halts pending /aw-setup)

```sh
for nested in aw aw-setup aw-tester-chrome; do
  [[ -f "$SKILL_DIR/$nested/SKILL.md" ]] || continue
  ln -sfn "$SKILL_DIR/$nested" "$CLAUDE_DIR/skills/$nested"
done
```

<sup>`pr-reviewer` · commit `8766952`</sup>



---

mthines (commented):
Reviewed `8766952` — 1 suggestion — no blocking issues; install.sh, l1.mjs and the conversion verified clean by execution. 1 finding(s) inline. This access path (GitHub MCP) has no issue-comment update tool, so the existing sticky report could not be rewritten in place.

---

skills/workflow/autonomous-workflow/install.sh
mthines: Fixed in `3d4be91`, slightly wider than the sketch.

This arrived at the review-loop's iteration cap, so it would normally have been surfaced rather than fixed. I applied it anyway because it isn't really a fifth round of the same argument: the block whose rationale you're quoting is one this PR wrote, and `aw-setup` is not a missing convenience but a dead end — `install.sh`'s own summary sends the user to `/aw-setup`, `aw-planner.agent.md:196` **halts** on a UI task until an aw-target exists, and `phase-1-planning.md:449` forbids the planner from scaffolding one itself. Three surfaces pointing at a skill that was never installed. The cap forbids another review iteration, not the fix; there is no sixth dispatch.

Two departures from your patch:

- **No `|| continue`.** A missing `SKILL.md` should fail, not silently skip — the existence guard added in `cc9a8c5` exists precisely so an incomplete skill directory errors instead of producing a dangling `/aw`. I generalized that guard into the same `NESTED_SKILLS` loop rather than letting the link loop swallow the case it was written to catch.
- **The dev-mode hop loops too.** Your patch links from `$SKILL_DIR` unconditionally, which would drop the cross-tool `~/.agents/skills/` tier for the two new entries in `--development` — `aw` has it and the siblings would not. All three now go through it, with the same `NESTED_LINK_BASE` fallback to `$SKILL_DIR`.

One thing your finding didn't reach, found while confirming it: `sync-symlinks.sh`'s `AW_SKILLS` bundle omitted both as well. The default recursive sync finds them (it walks for any dir with a `SKILL.md`), so only `--aw` was affected — same shape as the install.sh gap, different file.

Verified in a sandboxed `HOME` across all three modes with a legacy `aw.md` agent symlink pre-seeded: all three nested skills resolve, the legacy agent is still removed, all three agents still link, re-runs exit 0, and `--project` writes nothing into `HOME`. Removing `aw-setup/` from a copied skill directory now exits non-zero naming that path and creates no link. L1 1337/1337; `bash -n` clean on both scripts.

Not guarded by an L1 check. The honest reason is that the assertion I'd want — "every nested `SKILL.md` is linked by `install.sh`" — is one I'd be writing against the same loop in the same commit, and `G2c` earned its place by being derived from two independent sources. This one would have only one. Recording it here instead so the next person adding a nested skill has the sentence: **add it to `NESTED_SKILLS` in `install.sh` and to `AW_SKILLS` in `sync-symlinks.sh`.**

---

PR #177, opened 2026-09-05
feat(pr-reviewer): optional agent0_org on the Fix-with-Agent0 buttons

mthines:
## Why

A reader who belongs to several Dash0 organizations lands in whichever one Agent0 last resolved for them — not necessarily the one the PR's code lives in. `agent0_org` in the review config now points the fix buttons at the intended organization.

## What changed

- `agent0_org` is an optional review-config key: set it and both fix buttons get `&org=<slug>` appended; omit it (the default) and no `org` parameter is sent at all
- Threaded through `build-agent0-link.mjs` (`--org`), the run-level config resolution, and **both** button call sites — the same once-per-run value as `--env`
- A malformed slug never becomes a *different valid-looking* slug: the script rejects it outright, and the shell resolution matches the whole value in one pass rather than peeling it in stages
- This repo's own `.github/review.yaml` sets `agent0_org: dash0-development`

## How to verify

- `node scripts/eval/l1.mjs` — 1375/1375, including new guards `G31j`–`G31m`, `G32q`, `G37c`

## Notes for reviewers

Like `agent0_environment`, this key picks a destination and gates nothing — `agent0_fix_links` remains the only repo-wide switch. Review caught two truncation leaks in the shell resolution (`org#123` and `"org # 123"`, each peeling to `org`); the fix replaced the staged stripping with a single match rather than patching each stage, and deliberately left the shared `strip()` alone so the `agent0_fix_links` opt-out cannot fail open.

https://claude.ai/code/session_018vqEFrNqNC9BrMze5tc2jC


mthines:
### ⚪ 1 finding

Adds an optional `agent0_org` review-config key that appends `&org=<slug>` to both Fix-with-Agent0 deep links.

**Warnings:** 1 non-blocking finding

|---|---|---|
| Shell-side org slug is truncated at `#`, not rejected | `agents/shared/rules/review-config.md:391` | ⚪ low |

<details>

**Needs attention**

| Gate | Status | Details |
|---|---|---|
| Description vs. code | ✅ | The description matches the diff, including the L1 count and the reject-not-drop framing. |
| Documentation | ✅ | New § Organization section, schema comment, CLAUDE.md and README.md all updated in step. |
| Self-review signals | ✅ | No debug logs, leftover markers, or unreviewed stubs on new lines. |
| Code review | ⚠️ | 1 non-blocking finding — see the inline comment. |

**Found**

Quality — produced 7 → posted inline 1 · cleared 1 · carried forward 0 · deferred 0 · below-bar 2
Dropped — relevance 0 · dedupe 0 · grounding 0 · verifier-contradicted 2 · materiality 2 · shape 0
Severity — ⚪ 1 low
Verified — `node scripts/eval/l1.mjs` → 1368/1368, matching the PR body. The finding was reproduced by executing the `review-config.md` shell fragment and `build-agent0-link.mjs` on the same inputs.

**Run**

full · 253 lines in delta · tier deep · depth checkout · 9 files touched · blast_radius=low · semver_delta=none · high_stakes=0 · risky_shapes=none
⚠️ Write path is MCP-relayed, so the Fix-with-Agent0 buttons were withheld on both surfaces — every fix link exceeds the 140-char relay budget by construction (`agent0-fix-links.md` § Relay length limit).
CI — No checks have reported on `2c6892e` (combined status pending, 0 checks). Informational — CI never moves this verdict.
Memories — 8 indexed · 0 used

<sup>Nothing to report — standards (1 doc), optimality (skipped), measurability (skipped), integrations (not activated), 0 files skipped.</sup>

</details>

<sup>`pr-reviewer` · commit `2c6892e` · full review · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md) · updated <relative-time datetime="2026-09-05T08:20:00.000Z">Sep 5, 2026 8:20am UTC</relative-time></sup>

---

agents/shared/rules/review-config.md
```diff
@@ -373,13 +382,20 @@ if [ -n "$BASE_CONFIG_CONTENT" ]; then
     true)  AGENT0_FIX_LINKS="true" ;;
     false) AGENT0_FIX_LINKS="false" ;;
   esac
+  # DESTINATION ONLY, and OPTIONAL. Absent leaves AGENT0_ORG empty, and an empty AGENT0_ORG means
+  # the link carries no `org` parameter at all — not `org=`, not a default slug. Validated with the
+  # same pattern build-agent0-link.mjs enforces (agent0-fix-links.md § Organization), so a typo'd
+  # slug leaves the variable empty here instead of reaching the script as a value that would append
+  # a stray query parameter or break the button's href.
+  org=$(grep -E '^agent0_org:' <<< "$BASE_CONFIG_CONTENT" | strip)
+  if [[ "$org" =~ ^[A-Za-z0-9][A-Za-z0-9._-]{0,62}$ ]]; then AGENT0_ORG="$org"; fi
```
mthines: suggestion (low): ⚪ **Shell-side org slug is truncated at `#`, not rejected** **(non-blocking)**

`strip()` cuts at the first `#`, so `agent0_org: org#123` resolves to `org` — a valid-looking slug for a different organization, while the script rejects that same value.

Evidence: `agents/shared/rules/review-config.md:391` (strip() applies s/#.*$// first)

```diff
# strip(): anchor the comment cut to whitespace
-s/#.*$//
+s/[[:space:]]+#.*$//
```

<sup>`pr-reviewer` · commit `2c6892e`</sup>



---

mthines (commented):
---

agents/shared/rules/review-config.md
mthines: Addressed in `7342561` — confirmed and fixed, though not with the literal diff.

The defect reproduces exactly as described: `agent0_org: org#123` resolved to `org`, a valid-looking slug for a different organization, which is the failure mode the validation exists to prevent.

Anchoring the cut **inside `strip()`** would have traded it for a worse one. That helper is shared with `agent0_environment` and `agent0_fix_links`, and the anchored form stops `agent0_fix_links: false#x` from matching `false`, so it falls through to the default — the opt-out failing **open**:

```
agent0_fix_links: false#x    current: false     suggested: false#x  → no case match → ON
```

The two enum keys tolerate the loose helper because a mangled value fails their `case` and lands on a safe default; a free-form value has no backstop — it *becomes* the setting. So the fix is scoped to `agent0_org`, which now uses YAML's own comment rule (a `#` opens a comment only after whitespace), leaving `org#123` intact to be rejected while `dash0-development   # our org` still strips correctly.

Your point about the L1 gap was the useful half: `G31l` already carried a `#` case and `G32q`'s list did not, which is exactly how the divergence shipped. Both lists now cover it, plus a new check pinning that the fix stays scoped — verified to bite by reverting to the shared helper.

---

agents/shared/rules/review-config.md
```diff
@@ -373,13 +382,33 @@ if [ -n "$BASE_CONFIG_CONTENT" ]; then
...
+  # organization that sails through the check below. The two keys above can share `strip()` safely
+  # because a mangled value simply fails their `case` and falls through to a safe default; a
+  # free-form value has no such backstop — it BECOMES the setting. So `org` gets YAML's own comment
+  # rule (a `#` opens a comment only after whitespace), which leaves `org#123` intact to be
+  # rejected while still honouring the schema's documented ` # trailing comment` style.
+  #
+  # Anchoring the cut inside `strip()` itself would fix this key and break a more important one:
+  # `agent0_fix_links: false#x` would stop matching `false` and fall through to the default — the
+  # opt-out failing OPEN, the one direction this whole block exists to prevent.
+  org_strip() { sed -E 's/^[^:]+:[[:space:]]*//; s/[[:space:]]+#.*$//; s/["'"'"']//g; s/[[:space:]]*$//'; }
```
mthines: suggestion (low): ⚪ **Quoted org slug still truncated at a spaced #** **(non-blocking)**

`org_strip()` cuts the comment before stripping quotes, so `agent0_org: "org # 123"` resolves to `org` — accepted, opening a different organization than the config named.

Evidence: `agents/shared/rules/review-config.md:402` (executed; both accept org)

```diff
# org_strip(): a quoted scalar has no comment inside it — take it whole first
-s/[[:space:]]+#.*$//; s/["']//g
+s/^"([^"]*)".*$/\1/; t; s/^'([^']*)'.*$/\1/; t; s/[[:space:]]+#.*$//
```

<sup>`pr-reviewer` · commit `b32e371`</sup>



---

mthines (commented):
---

agents/shared/rules/review-config.md
mthines: Addressed in `3f544fe` — reproduced exactly (`agent0_org: "org # 123"` → `org`), and you're right that line 402's own comment made quoting look safe.

Rather than patch the third stage, I took the finding as being about the *approach*. Peeling a value in stages means every stage can hand back a prefix of the configured text — `org#123` → `org` was stage one, `"org # 123"` → `org` is stage two, and a quote-stripping fix would just move the seam again. So the key now asserts the whole value in one match:

```bash
^agent0_org:[[:space:]]*([A-Za-z0-9][A-Za-z0-9._-]{0,62})[[:space:]]*([[:space:]]#.*)?$
```

A bare slug, optionally followed by a YAML comment. Anything else matches nothing and leaves `AGENT0_ORG` empty — the fail-safe state. That deletes `org_strip()` instead of growing it, and the shared `strip()` stays untouched so the `agent0_fix_links` opt-out still can't fail open.

One accepted cost, stated in the code: `agent0_org: "dash0-development"` is valid YAML and is now rejected. That yields an absent parameter, never a wrong organization, and the schema documents the unquoted form.

Four cases added to `G32q` — both quote styles, an embedded-quote form, and `"org#123"` — verified to bite by restoring the old helper. 1375/1375.

---

PR #180, opened 2026-09-06
Test dispatch availability as a capability, not a tool name

mthines:
## Summary

A Full `aw` run in a Claude Code cloud session delivered a draft PR flagged `NOT REVIEWED`, and the review-recovery path fired and recovered nothing. Probing this harness — rather than reasoning about it — turned up three defects. A fourth came out of this PR's own review.

**1. The capability gates were tool-NAME checks.** `review-loop`'s Step 0 precondition read *"check whether `Task` appears in your available tools"*, and `pr-review` and `aw` carried the same literal in eight more places. The Claude Agent SDK harness behind Claude Code on the web and in cloud sessions **holds** the capability and spells the tool `Agent`, so the gate concluded "no dispatch available" and skipped the review on a PR that was fully reviewable.

The failure hid itself: a skip is a documented legitimate outcome, so nothing downstream contradicted it. Every check now asks whether *any* available tool dispatches a sub-agent — a tool taking a `subagent_type` (or equivalent) parameter is the dispatch tool whatever it is called — and each file states the name check as a `WRONG`/`RIGHT` example so it reads as a defect rather than as the rule.

**Dispatch gates changed in five skills, not three.** Beyond `autonomous-workflow`, `review-loop`, and `pr-review`, the same defect was live in two more and is fixed here:

| Skill | Site |
| --- | --- |
| `autonomous-workflow` | `templates/routing.rule.md` (the **installed** rule, symlinked into `~/.claude/rules/`), `SKILL.md`, `CLAUDE.md` (design-intent invariant + smoke-test step) |
| `review-loop` | `SKILL.md` Step 0 precondition + 8 downstream sites |
| `pr-review` | `SKILL.md` Step 2 + hard rules |
| `review-changes` | `SKILL.md` — the `--external-review` row |
| `preview-spec` | `rules/runner.md` — four Playwright-driver availability gates |

Left deliberately untouched: every `WRONG`/✗ anti-pattern block, the illustrative `Task(subagent_type=…)` dispatch calls, and the v3.19.0 history entry (it describes what was true then).

[...]

PR #181, opened 2026-09-06
Make the preview-spec author delegation reachable and its skip visible

mthines:
## Why

Preview-spec blocks were almost never reaching PR descriptions. Measured rather than assumed: of the **20 most recent lorekit PRs, all 20 merged after preview-spec shipped (2026-09-01) and none carried a `preview-spec:v1` block** — including #651, whose diff matches `create-pr` Step 6.4's own UI heuristic on nine files. Two independent defects produced that, and either alone is sufficient.

## What changed

**1. `preview-spec`'s grant could not reach GitHub on the mcp path.** Its frontmatter listed `Bash(gh *)` and no `mcp__github__*`, while its own Step 0 told it to resolve to gh / mcp / neither and its Step 4 deliverable was `gh pr edit --body`. `gh` is absent in cloud sessions, which is where these runs happen, so the operation's only deliverable was unreachable. The grant also named `Task` alone — the same literal-spelling defect #180 removed from this skill's *prose*, surviving in the frontmatter that commit never touched — so `run` could not dispatch `aw-tester` under the `Agent` spelling. Step 0 now names the mcp equivalent of each `gh` command, because citing `github-access.md` and then printing only `gh` forms leaves the reader to invent the call.

`run`'s deployment lookup genuinely has no mcp equivalent (no `mcp__github__*` tool exposes deployments), so that path now requires an explicit `--url` and reports `inconclusive: no access path for deployment lookup (pass --url)`, never `preview not deployed` — which asserts a fact about a deployment nothing checked. That branch is enforced in `preview-url-resolution.md`, the file that **owns** the resolution outcome, because `runner.md § Step 2` delegates resolution there and declares any `inconclusive` from there terminal — a rule stated only in `SKILL.md` Step 0 is unreachable by a conforming run, and Step 0 cannot see the branch's condition anyway (*`run` invoked without `--url`* is an argument, resolved after the access path).

**2. `create-pr` Step 10 had no slot for the spec**, so all five skip conditions and the failure mode rendered as a clean, successful PR. That is why defect 1 survived four days of UI PRs unnoticed — the `F6`/`F7` shape, where a degraded path reports as a legitimate outcome. The line is mandatory on every run, non-UI diffs included, since `not authored (no UI files in diff)` is the informative answer there.

**3. `review-loop` Step 1.6's outcome table** dropped four of `run`'s returns on the floor and mapped a fifth to advice that cannot come true. Its `preview not deployed` row said *re-run once the preview is up* — unactionable when no lookup ran — and `empty spec` plus the three distinct `NOT RUN (…)` strings had no row at all. `empty spec` gets its own row rather than folding into `no spec`: markers present with an empty body means `author` ran and embedded nothing, a spec-authoring bug, where `no spec` legitimately covers "this PR needed none". A final catch-all quotes an unrecognised return verbatim, since a delegate gaining an outcome this table has no row for is exactly how the permanently-false note reached a report.

**4. `G49`** asserts all of the above, plus the run-path degradation and each report outcome.

**5. `G49-lint`** makes the recurring guard defect below structurally unrepresentable: it reads `l1.mjs`, derives the `G49` block's whole-file `read()` bindings mechanically, and requires every `.test()` against one of them to be **line-anchored** (`^` in the body, `m` in the flags), which pins the match to one line or one table row. It is fail-closed — a receiver it cannot parse as a regex literal is a violation, not an exemption, since unparseable is the hole a hoisted pattern constant or a `new RegExp(...)` would slip through. Three self-sentinels stop the lint from passing vacuously itself (block located and large, ≥ 3 bindings derived, ≥ 10 `.test()` calls seen); there is deliberately **no** floor on the whole-file count, because that is the quantity the rule drives to zero and a floor would redden on exactly that improvement.

The lint ships with **no exemptions**, which required re-pointing the block's remaining bare whole-file assertions first: Step 0's two access-path checks, the access-path precondition pair, and `create-pr`'s Step 6.4 / Step 10 checks now read section slices through a local `section()` wrapper over the shared `extractSection`, each proved non-empty by its own sentinel; the two surviving `RL` checks are row-anchored, which also removed the last character-gap-as-scope (`{0,320}`) in the block.

**6. Two real defects in the L2 harness, found while diagnosing the red `l2` gate.** Neither fixes the gate — see *Stated limitations* — but both are independently correct and both are the same self-concealing shape as the rest of this PR.

- **The `code-review-retrieval-relevance` suite was fed a rubric 2.5× broader than its own question.** Its `instruction` names Step 1.0 + Step 1.2c; `rubric.section: "## Step 1"` fed all ten `### 1.x` subsections (67,630 chars vs. 27,568), including `### 1.2d`, whose diff-keyed shortlist answers *what reaches the finders* rather than *what the documented read returns* — so a list-reachable lesson unrelated to the diff was legitimately `skip` under the rubric and the suite contradicted its own `instruction`. `l2.mjs` gains a `rubric.sections` list form resolved by a fail-closed `rubricFor()` (empty list, both keys set, or an unrecognised shape is an error, never an empty prompt), and `G21a` now asserts both subsection literals **and** the absence of the parent, so a silent re-widening reds L1 rather than only the paid L2 run.
- **`### 1.0` never stated what its read does *not* filter on.** The four `memory_list` calls filter on tag, scope and expiry and nothing else — no diff parameter, no `seen_count` — but the prose said so nowhere, leaving a reviewer free to narrow the read by its own sense of relevance. Now stated prescriptively, with where diff-keying and `seen_count` actually enter (Step 1.2c/1.2d additively; promotion and Step 2.7b respectively).

Three quiet-failure holes in the harness closed alongside them, all instances of this PR's class:

| Hole | Fix |
|---|---|
| `--suite <typo>` matched nothing, graded **zero cases**, exited 0 | exits 1 and names the suites |
| `G21g`'s `>= 7` floor let a suite with an unrecognised `rubric:` shape go unchecked while green | parity check against the suite count |
| the `section:` parse silently rejected a trailing comma — which the floor would have hidden | `,?` tolerated, so the guard names the real fault |

`evals-l2.yml` also gains a `workflow_dispatch` `suite` input: iterating on one suite was a ~6-minute nine-suite run per attempt, all billed to the same key.

## Verification

- **L1 1474 → 1518** at every commit: +44 checks, 23 of them named `G49 …`, 8 named `G49-lint …` (3 sentinels + 5 per-assertion), 1 the new `G21g` parity sentinel, the rest from the per-outcome loops. `l1` is **green** on `d72d196`; `l2` is **red** there and identically red on `main` — see *Stated limitations*.
- **`G49` bite-tested against every fix**, each mutation's expectation declared *before* the run so a green cannot be reinterpreted as a pass afterwards. Deletion tests want RED; meaning-preserving edits want GREEN. Latest matrix:

  | Mutation | Result | Wanted |
  |---|---|---|
  | delegation moved out of Step 2 into a promoted `# ` heading | RED | RED |
  | delegation sentence deleted | RED | RED |
  | terminality sentence removed | RED | RED |
  | Step 2 heading renamed | RED | RED |
  | Step 2 demoted to `###` | RED | RED |
  | catch-all row inverted to "treat as green" | RED | RED |
  | catch-all row deleted | RED | RED |
  | `unrecognised outcome` dropped from the report slot | RED | RED |
  | Step 1 reverted to the gh-only body read | RED | RED |
  | grant dropped but still named in `description:` | RED | RED |
  | trailing space on the Step 2 heading | GREEN | GREEN |
  | bolding the catch-all cell | GREEN | GREEN |
  | fenced `## ` line inside Step 2 | GREEN | GREEN |
  | reflow before the terminality claim | GREEN | GREEN |
  | swapping two arms of an unordered enumeration | GREEN | GREEN |

- **`G49-lint` bite-tested the same way**, five mutations, all expectations declared first, all wanting RED and all RED — with the failure text naming the exact `l1.mjs:<line> .test(<binding>)`:

  | Mutation | Result | Wanted |
  |---|---|---|
  | a re-pointed check reverted to a bare `.test(PS)` | RED | RED |
  | `^` and the `m` flag dropped from a row-anchored `.test(RL)` | RED | RED |
  | receiver replaced with `new RegExp(…, "m")` (not a literal) | RED | RED |
  | the `G49` banner renamed, so the block cannot be located | RED | RED |
  | the binding-derivation regex broken, so zero bindings resolve | RED | RED |

  The last two are the self-sentinels: both red by name *and* the check total drops by 5, because the per-assertion checks vanish when the scan finds nothing.

- **The three L2-harness guards bite-tested, expectations declared first, all RED as declared:** renaming the `sections` key → parity sentinel red naming `suites=9 parsed=8`; widening the rubric back to `## Step 1` → `G21a` red; a broken subsection literal → `G21g` non-empty-body red quoting `section "### 1.2c Nonexistent heading" not found`.
- **The other eight L2 suites are unaffected** by `rubricFor` — all clear the 70% floor on the nine-suite run ([34044749700](https://github.com/mthines/agent-skills/actions/runs/34044749700)): `bug-class` 100%, `aw-should-trigger` 100%, `optimize-approach-optimality` 100%, `reviewer-agreement-bump` 100%, `severity-tiering` 93.3%, `complexity-triage` 92.9%, `tier-routing` 83.3%, `shape-depth-routing` 81.8%.
- **Behavioural A/B, 3 rounds**: pre-fix vs post-fix text on a no-`gh` inventory. Post-fix **9/9 cells across 3 rounds** with byte-identical report lines; pre-fix answered `NO SLOT` 3/3.

## What the review found, and why it took five passes

**Eight** of the guards in this PR were **vacuous on arrival** — five of them found inside the fix for the previous one, and the eighth found by the lint's own logic while enumerating the block *before the lint existed*. They are all one defect class: *a substring test whose satisfying occurrence is not the one the check names.*

| # | The check | What satisfied it instead |
|---|---|---|
| 1 | `!Task \|\| Agent` | neither holding — the disjunct is satisfied by absence |
| 2 | bare `"authored"` | `"not authored"`, the decline outcome's own text |
| 3 | `slice(0, 400)` on a 253-char section | 147 chars of the *next* section |
| 4 | `empty preview-spec block` | the report slot, which enumerates the same string |
| 5 | `unrecognised outcome` | same — and the row could be **inverted**, not just deleted, with L1 silent |
| 6 | `indexOf` on a heading's bare text | the same text inside a *deeper* heading |
| 7 | `indexOf("\n## ")` as an end bound | a *promoted* next heading, growing the slice 253 → 1813 |
| 8 | `/pass --url/` over `preview-spec/SKILL.md` | Operation `run`'s own `--url` prose, 30 lines below the Step 0 outcome string the check names |

Two structural lessons are now recorded in the code rather than only in the thread history:

- **A `length > N` sentinel is a lower bound and cannot see an over-capture.** It is what kept #3, #6 and #7 invisible; #3's fix reintroduced the same blindness at the other end of the slice.
- **A guard can rest on an unguarded premise.** The block argued reachability *from* `runner.md` in two comments and read that file in neither, so deleting either load-bearing sentence left L1 green with the branch unreachable again.

The seventh fix removes the private slicer entirely in favour of the shared, level- and fence-aware `extractSection`, and fixes *its* start bound — the same bare `indexOf` — in `lib.mjs` rather than working around it. `G21g` already drives all seven L2 rubric headings through that helper, so the change lands with existing coverage.

Three checks that could only ever **false-red** were also removed: a 400-char section window, an 80-char gap that used 18 of its 80, and two order- and decoration-pinned matches. A guard that reddens on a reflow trains the next author to stop reading it.

## Stated limitations

- **`l2` is red at merge, at 3/5 (60.0%) on the `code-review-retrieval-relevance` suite, and item 6 above did not fix it.** Both hypotheses were falsified against CI: the rubric narrowing ([34044758211](https://github.com/mthines/agent-skills/actions/runs/34044758211)) and the `### 1.0` negatives ([34044924487](https://github.com/mthines/agent-skills/actions/runs/34044924487)) each returned an identical 3/5 with the same two misses. It is **not this PR's failure** — `main` at `b6fee49` fails identically ([34041417796](https://github.com/mthines/agent-skills/actions/runs/34041417796)) — and it is a **first execution, not a regression**: the `l2` step ran 0 s on `1882840` (the clean `no ANTHROPIC_API_KEY — skipping` exit, treated as a pass by design) and ~6 min once the secret landed, so every prior "l2 green" on this PR was a skip. That is the same `F6` shape this PR is about, one layer up in the harness.
- **The remaining half of that fix is deliberately deferred to a follow-up PR.** The one untouched candidate is the suite's `instruction` string, whose "surfaced by the documented read **for the given PR diff**" clause pragmatically frames the task as a relevance judgment. Fixing it alone is not defensible: the golden set is 4 `surface` / 1 `skip`, so an always-`surface` responder already scores 80% and clears the floor — disambiguating without adding genuine `skip` decoys (wrong scope, expired record, wrong tag, other-repo `repo::`) would be handing over the answer key. Neither failing record even states a `scope`. So the honest fix is a pair — instruction rewrite **plus** an expanded, scope-specified golden set — on a set stamped `BOOTSTRAP SEED — NOT A REAL BASELINE`, which is a design task rather than a CI-fix commit. No golden case was edited, skipped or added here, and `EVAL_GATE` was not touched.
- The behavioural probe did **not** reproduce defect 1. Handed its tool inventory, the pre-fix arm bridged the missing mapping itself. `allowed-tools` is a permission allowlist, so a missing tool is uncallable however well the agent reasons — but no prompt-level simulation exercises permission enforcement. Defect 2 is the half with measured behavioural support; defect 1 rests on the allowlist contract. Recorded in `G49`'s own comment so a later reader doesn't infer equal support.
- The review loop **hit its 5-iteration cap** rather than converging on a quiet pass. Every thread is resolved, but the last two iterations each found a real defect inside the previous iteration's fix, so "no further findings" has not been demonstrated — only "no findings left open".
- **`G49-lint` is scoped to the `G49` block alone**, located by its exact banner. Other blocks in `l1.mjs` still assert against whole files unlinted; generalising the rule to the whole file is a separate change with its own baseline to establish.

## Not in this diff

The Step 6.4 heuristic itself is unchanged. Its "test-only and config-only changes do not count" sentence could plausibly make an agent decline a diff whose only matches are `.stories.tsx` files, but #651 also changed five plain components, so that wording is not what failed here and narrowing it would be a guess.

https://claude.ai/code/session_0134kyioxSrXMMz4KwZDm7QR


mthines:
### 🟡 5 findings

Closes four vacuous-guard defects in G49 and moves the no-access-path branch into the file that owns URL resolution. Three of the four guard fixes bite; one is inert.

**Warnings:** stale verification counts in the PR body; three G49 assertions still inert or vacuously passable

|---|---|---|
| anchor rationale is false and the fix is a no-op | [`scripts/eval/l1.mjs:5425`](https://github.com/mthines/agent-skills/blob/882bcef3ac4833d94fedc03278948a454e683c1d/scripts/eval/l1.mjs#L5425) | 🟡 medium |
| two grant assertions still read the whole frontmatter | [`scripts/eval/l1.mjs:5342`](https://github.com/mthines/agent-skills/blob/882bcef3ac4833d94fedc03278948a454e683c1d/scripts/eval/l1.mjs#L5342) | 🟡 medium |
| reachability premise rests on an unasserted third file | [`scripts/eval/l1.mjs:5393`](https://github.com/mthines/agent-skills/blob/882bcef3ac4833d94fedc03278948a454e683c1d/scripts/eval/l1.mjs#L5393) | 🟡 medium |
| gh-only body read in the runner's own Step 1 | [`skills/testing/preview-spec/SKILL.md:98`](https://github.com/mthines/agent-skills/blob/882bcef3ac4833d94fedc03278948a454e683c1d/skills/testing/preview-spec/SKILL.md#L98) | ⚪ low |
| catch-all covers only the inconclusive family | [`skills/quality/review-loop/SKILL.md:663`](https://github.com/mthines/agent-skills/blob/882bcef3ac4833d94fedc03278948a454e683c1d/skills/quality/review-loop/SKILL.md#L663) | ⚪ low |

<details>

**Needs attention**

| Gate | Status | Details |
|---|---|---|
| Description vs. code | ⚠️ | Verification says L1 `1488/1488` (14 new); HEAD is `1496/1496` (22 new), and `bite-tested 8/8` fails for one fix. |
| Prior review feedback | ✅ | no review threads or issue comments on this PR |
| Documentation | ✅ | each fix carries its rationale in the file it lands in; inventory prose in `CLAUDE.md` matches the code |
| Self-review signals | ✅ | no debug output, unfinished markers, suppressions, or conflict residue on new lines |
| Code review | ⚠️ | 5 non-blocking findings — 3 guard-integrity gaps in the new G49 block, 2 in the prose it guards |

**Found**

Quality — produced 9 → cleared 5 → posted inline 5; 4 contradicted at verification and dropped; 0 deferred over cap, 0 suppressed by memory
Severity — 🟡 3 medium · ⚪ 2 low
Verified — each of the four claimed guard fixes bite-tested by reverting it and re-running L1 — 3 go red, 1 stays green (finding 1); 2 new vacuity holes found the same way

**Run**

full · 252 lines in delta · tier deep · depth checkout · re-review of the 882bcef head; prior pass reported in-context, so no GitHub comment history to dedup against
CI — `evals-l1` green on `882bcef`; no other check runs on this diff
Memories — 16 indexed · 0 used

<sup>Nothing to report — standards (3 docs), optimality (4 judged), measurability (skipped), integrations (not activated), 0 files skipped.</sup>

</details>

<sup>`pr-reviewer` · commit `882bcef` · full review · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md) · updated <relative-time datetime="2026-09-06T12:20:09.000Z">Sep 6, 2026 12:20pm UTC</relative-time></sup>

---

scripts/eval/l1.mjs
```diff
@@ -5298,4 +5298,162 @@ const isPollBlock = (block) =>
...
+      + " even a correctly-behaving `run` fell through the outcome table unmapped");
+  s.check("G49 review-loop's remedy for it is an explicit URL, not waiting for a build",
+    /no deployment lookup on this access path[\s\S]{0,320}--url <preview-url>/.test(RL),
+    "`re-run once the preview is up` is unactionable here — nothing was looked up, so the"
+      + " human re-runs and gets the same string; only passing a URL changes the outcome");
+
+  // 4. The report slot. Every skip condition Step 6.4 enumerates needs a rendered outcome,
+  //    or the degraded path reports as success.
+  // Anchored at line start, because the report line IS a whole line while the string
+  // `Preview spec (Step 6.4):` also occurs mid-sentence in Step 6.4's own prose — an
```
mthines: issue (medium): 🟡 **anchor rationale is false and the fix is a no-op** **(non-blocking)**

`Preview spec (Step 6.4):` occurs once in create-pr — the slot itself — so both regexes match the same index and reverting the anchor keeps L1 green. Anchoring is right; the reason given is not.

Evidence: `skills/delivery/create-pr/SKILL.md:348` (the only occurrence) · `scripts/eval/l1.mjs:5428` (both match index 24161)

<sup>`pr-reviewer` · commit `882bcef`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -5298,4 +5298,162 @@ const isPollBlock = (block) =>
...
+  // The sweep must be proved non-empty before any assertion over it is trusted.
+  s.check("G49 the guard reads both surfaces",
+    PS.length > 2000 && CP.length > 2000 && PS_FM.includes("allowed-tools:"),
+    "an empty read or a frontmatter slice that missed `allowed-tools:` would pass every"
+      + " assertion below vacuously");
+
+  // 1. The grant must be able to reach GitHub on BOTH access paths, and to dispatch under
+  //    either spelling. `Bash(gh *)` alone is the pre-fix state; `Task` alone is the exact
+  //    literal-name defect PR #180 removed from this skill's PROSE while leaving it here.
+  s.check("G49 preview-spec's grant can write a PR body on the mcp path",
```
mthines, line 5405: issue (medium): 🟡 **two grant assertions still read the whole frontmatter** **(non-blocking)**

Both test `PS_FM`, so a tool named in `description:` prose but absent from `allowed-tools:` passes — bite-tested green. Line 5349 states the rule they break; `PS_TOOLS` on 5352 is the fix.

Evidence: `scripts/eval/l1.mjs:5349` (the rule stated) · `scripts/eval/l1.mjs:5352` (PS_TOOLS already exists)

```diff
-    /mcp__github__update_pull_request/.test(PS_FM),
+    /mcp__github__update_pull_request/.test(PS_TOOLS),
-    /mcp__github__pull_request_read/.test(PS_FM),
+    /mcp__github__pull_request_read/.test(PS_TOOLS),
# and move both s.check calls below the PS_TOOLS binding
```

<sup>`pr-reviewer` · commit `882bcef`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -5298,4 +5298,162 @@ const isPollBlock = (block) =>
...
+  //    must be stated rather than collapsed into the deployed/not-deployed verdict.
+  s.check("G49 preview-spec distinguishes 'no access path' from 'preview not deployed'",
+    /pass --url/.test(PS) && /never report `inconclusive: preview not deployed`/.test(PS),
+    "no `mcp__github__*` tool exposes deployments; reporting a lookup that never happened"
+      + " as `preview not deployed` asserts a fact about the deployment as if checked");
+
+  // 3b. And it must bite in the file that OWNS the decision. Asserting it against SKILL.md
+  //     alone was structurally incapable of catching the residue: `runner.md § Step 2`
+  //     delegates resolution to `preview-url-resolution.md` and declares any `inconclusive`
+  //     from THERE terminal, so an agent following the file the skill points at reached the
```
mthines, line 5458: issue (medium): 🟡 **reachability premise rests on an unasserted third file** **(non-blocking)**

True of `runner.md` today, but no G49 check reads that file. Deleting its `terminal for this run` sentence, or inlining its lookup, leaves L1 at 1496/1496 while this precondition stops being honoured.

Evidence: `skills/testing/preview-spec/rules/runner.md:32` (the terminal sentence) · `skills/testing/preview-spec/rules/runner.md:31` (the delegation)

<sup>`pr-reviewer` · commit `882bcef`</sup>



---

skills/testing/preview-spec/SKILL.md
```diff
@@ -90,6 +90,23 @@ Resolve which path you have — `gh` CLI, `mcp__github__*` tools, or neither —
 Resolve once, state the path, and use it for the whole run.
 The commands below are the `gh`-path form.
 
+**On the `mcp` path, use these equivalents.**
+Naming them here is load-bearing: the `gh`-path form above is not a mapping, and a reader who has to invent one writes nothing to the PR.
+
+| `gh`-path command | `mcp`-path equivalent |
+| --- | --- |
+| `gh pr view <pr> --json body` | `mcp__github__pull_request_read` with `method: "get"` |
```
mthines, line 98: suggestion (low): ⚪ **gh-only body read in the runner's own Step 1** **(non-blocking)**

`run`'s procedure file gives the body read as `gh pr view` with no mcp form and no pointer here, and its Step 1 precedes the Step 2 precondition this PR adds — same residue, one layer lower.

Evidence: `skills/testing/preview-spec/rules/runner.md:19` (gh-only body read)

<sup>`pr-reviewer` · commit `882bcef`</sup>



---

skills/quality/review-loop/SKILL.md
```diff
@@ -649,6 +659,8 @@ Map its outcome into the report:
 | --- | --- |
 | `no spec` (no block — not a UI PR, or `author` never ran) | `not run (no preview-spec block)` — log and continue |
 | `inconclusive: preview not deployed` | `inconclusive (preview not deployed at exit)` — note `re-run /preview-spec run <PR-URL> once the preview is up`. Never a red |
+| `inconclusive: no access path for deployment lookup (pass --url)` | `inconclusive (no deployment lookup on this access path)` — note `re-run /preview-spec run <PR-URL> --url <preview-url>`. Never a red, and never recorded as `preview not deployed`: no lookup ran, so waiting for the build fixes nothing and only an explicit URL changes the outcome |
+| any other `inconclusive: <reason>` (`preview building`, `no preview environment`, `preview deploy failed`, `preview URL not published`) | `inconclusive (<reason> at exit)` — log the reason verbatim and continue. Never a red |
```
mthines, line 663: suggestion (low): ⚪ **catch-all covers only the inconclusive family** **(non-blocking)**

`run` also returns `empty spec` and `NOT RUN (chrome unavailable, user declined Playwright)`; neither is an `inconclusive:`, a `no spec`, a verdict, or a missing skill, so both still fall through.

Evidence: `skills/testing/preview-spec/rules/runner.md:23` (empty spec) · `skills/testing/preview-spec/rules/runner.md:54` (NOT RUN declined)

<sup>`pr-reviewer` · commit `882bcef`</sup>



---

mthines (commented):
---

scripts/eval/l1.mjs
mthines: Confirmed and fixed in `f58a3c4`. `grep -n 'Preview spec (Step 6\.4):'` returns exactly one line (348), so the rationale was describing a state that no longer exists: an earlier revision of Step 6.4's prose *did* repeat the literal, which is what broke the unanchored `exec` and turned L1 red — I reworded that prose rather than leaving the guard dependent on the wording, and then left the comment describing the old world.

Kept the anchor and the arm count, both now documented as defensive rather than load-bearing, with the reason they stay: they cost nothing and they are what stops a re-added prose mention from silently re-breaking the capture.

The PR body's bite-test tally is also wrong on this point and gets corrected in the description refresh — it credited a fix that does not bite.

---

scripts/eval/l1.mjs
mthines: Confirmed and fixed in `f58a3c4`. This was the worst of the five — the comment on 5349 stated the exact rule the two assertions beneath it broke.

Fixed by moving the `PS_TOOLS` binding and its non-empty sentinel **above** both assertions rather than only swapping the operand, since binding it after the checks that depend on it is what let the mistake look plausible.

Bite-tested against the precise case you described — drop `mcp__github__update_pull_request` from `allowed-tools:` while adding it to `description:` prose:

```
BITE  F2/grant-dropped-but-in-description
BITE  F2/grant-dropped-entirely
```

Both red now; the first was green before.

---

scripts/eval/l1.mjs
mthines: Confirmed and fixed in `f58a3c4`. The block argued reachability *from* `runner.md` in two comments and read the file in neither — a guard resting on an unguarded premise, which is the same shape as the residue this PR was opened to fix.

Added two checks scoped to `runner.md § Step 2` (sliced from the heading, so a sentence elsewhere in the file cannot satisfy them): the delegation to `preview-url-resolution.md`, and the `terminal for this run` declaration. Both bite:

```
BITE  F3/step2-delegation-removed
BITE  F3/step2-terminality-removed
```

The second is the one that matters most — without terminality the runner can treat `no access path for deployment lookup` as a soft miss and carry on to dispatch, reporting a pass or a fail for a spec that never ran, which is the outcome the string exists to prevent.

L1 1496 → 1498.

---

skills/testing/preview-spec/SKILL.md
mthines: Confirmed and fixed in `88b3d5b`. `runner.md` Step 1 now names both calls (`gh pr view … --json body` / `mcp__github__pull_request_read` with `method: "get"`) and points at the Step 0 table that owns the mapping.

Added one sentence stating the contrast with Step 2 explicitly, because the asymmetry is the whole point and is otherwise easy to over-generalise: this read *does* have an mcp equivalent, so an absent `gh` never blocks it — unlike the deployment lookup two steps later, which has none.

---

skills/quality/review-loop/SKILL.md
mthines: Confirmed and fixed in `88b3d5b`. Five returns fell through, not two — `empty spec` plus all four `NOT RUN (…)` strings (`runner.md:54` twice, `:83`, `:94`).

`empty spec` gets its own row rather than folding into `no spec`: the markers existing with an empty body means `author` ran and embedded nothing, which is a spec-authoring bug worth naming, whereas `no spec` legitimately covers "this PR needed none". Collapsing them would report the bug as the healthy case.

The `NOT RUN (…)` family maps verbatim and is explicitly never a red — no driver executed, so there is no verdict to be red about.

Then a final catch-all row that quotes an unrecognised return verbatim and records it as neither a pass nor a skip, since a delegate gaining an outcome this table has no row for is precisely how the permanently-false `re-run once the preview is up` note reached a report in the first place. The Step 3 report slot enumerates the new arms too.

---

scripts/eval/l1.mjs
```diff
@@ -5298,4 +5298,193 @@ const isPollBlock = (block) =>
...
+      && PUR.indexOf("## The access-path precondition") < PUR.indexOf("## Resolution steps"),
+    "a precondition printed after the steps it gates is read too late; the condition is"
+      + " `run` invoked without `--url`, which SKILL.md Step 0 cannot see at all");
+  // And the premise that makes the two checks above MEAN anything must itself be guarded.
+  // Placing the branch in the owning file is only sufficient because `runner.md § Step 2`
+  // (a) delegates resolution there and (b) declares any `inconclusive` from there terminal.
+  // Neither was asserted, so deleting either sentence from `runner.md` left L1 green while
+  // the branch became unreachable again — the guard resting on an unguarded premise.
+  const RUN = read("skills/testing/preview-spec/rules/runner.md");
+  const RUN_STEP2 = RUN.slice(RUN.indexOf("## Step 2: Resolve the preview URL"));
```
mthines: issue (medium): 🟡 **Step 2 slice bleeds 147 chars into Step 3** **(non-blocking)**

Step 2's section is 253 chars but the window is 400, so Step 3's opening sits inside it. Deleting the delegation sentence and naming the file in Step 3 keeps L1 at 1498/1498.

Evidence: `skills/testing/preview-spec/rules/runner.md:30` (section 253 chars, window 400) · `scripts/eval/l1.mjs:5422` (bite: V2 green, V3 red)

```js
-const RUN_STEP2 = RUN.slice(RUN.indexOf("## Step 2: Resolve the preview URL"));
+const H = "## Step 2: Resolve the preview URL";
+const i = RUN.indexOf(H);
+const RUN_STEP2 = i === -1 ? "" : RUN.slice(i, RUN.indexOf("\n## ", i + 1));
# then assert against RUN_STEP2 whole — drop both .slice(0, 400)
```

<sup>`pr-reviewer` · commit `2b1df0a`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -5298,4 +5298,193 @@ const isPollBlock = (block) =>
...
+    /`inconclusive: ?…`[^\n]*terminal for this run/.test(RUN_STEP2.slice(0, 400)),
+    "without terminality the runner may treat `no access path for deployment lookup` as a"
+      + " soft miss and carry on to dispatch, reporting a pass or a fail for a spec that"
+      + " never ran — the outcome the string exists to prevent");
+
+  // 3c. The named downstream consumer must map the new outcome, and must NOT map it to the
+  //     other one's remedy. `review-loop` had a row for `preview not deployed` whose note
+  //     is `re-run once the preview is up` — advice that can never come true when no lookup
+  //     ran, so a UI PR reviewed on a cloud session got a permanently-false note in it.
+  const RL = read("skills/quality/review-loop/SKILL.md");
```
mthines, line 5563: suggestion (medium): 🟡 **three of this pass's five fixes landed unguarded** **(non-blocking)**

Reverting `runner.md` Step 1's dual path, review-loop's catch-all row, or its `empty spec` row each leaves L1 at 1498/1498. Step 2's premise got a guard this commit; these three did not.

Evidence: `skills/testing/preview-spec/rules/runner.md:18` (U1 revert: green) · `skills/quality/review-loop/SKILL.md:669` (U2/U3 revert: green)

<sup>`pr-reviewer` · commit `2b1df0a`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -5298,4 +5298,193 @@ const isPollBlock = (block) =>
...
+  const RUN = read("skills/testing/preview-spec/rules/runner.md");
+  const RUN_STEP2 = RUN.slice(RUN.indexOf("## Step 2: Resolve the preview URL"));
+  s.check("G49 runner.md Step 2 delegates resolution to the owning file",
+    RUN_STEP2.startsWith("## Step 2: Resolve the preview URL")
+      && /preview-url-resolution\.md/.test(RUN_STEP2.slice(0, 400)),
+    "the precondition is enforced in `preview-url-resolution.md`; a Step 2 that resolves the"
+      + " URL itself never reads that file, and the branch is unreachable however well it"
+      + " is written");
+  s.check("G49 runner.md Step 2 declares an inconclusive from that file terminal",
+    /`inconclusive: ?…`[^\n]*terminal for this run/.test(RUN_STEP2.slice(0, 400)),
```
mthines: suggestion (low): ⚪ **terminality guard asserts the ellipsis, not the rule** **(non-blocking)**

The pattern needs the literal `…`, so rewriting the sentence to `inconclusive: <reason>` turns L1 red while the rule stays honoured — the wording dependency this block's own comment disavows.

Evidence: `scripts/eval/l1.mjs:5430` (bite V4: reworded, still red)

```js
-/`inconclusive: ?…`[^\n]*terminal for this run/.test(RUN_STEP2)
+/inconclusive[^\n]{0,40}terminal for this run/.test(RUN_STEP2)
```

<sup>`pr-reviewer` · commit `2b1df0a`</sup>



---

CLAUDE.md
```diff
@@ -65,7 +65,7 @@ Type markers (by primary entry point — all three are technically model-invocab
...
...rface), and its outcome table had a row mapping `preview not deployed` to `re-run once the preview is up` — advice that can never come true when no lookup ran, so every UI PR reviewed on a cloud session got a permanently-false note in its report. It now maps the new outcome to `re-run … --url <preview-url>`, carries a row for the four other `inconclusive` reasons, and covers the five returns that previously fell through it entirely — `empty spec` on its own row (the block existing but empty means `author` ran and embedded nothing, a bug to name rather than a PR that needed no spec) and the four `NOT RUN (…)` driver-unavailable strings — behind a final catch-all that quotes an unrecognised return verbatim instead of guessing, an unmapped return being exactly how the permanently-false note got there. **Two-way self-improvement:** the runner writes navigation / spec-quality friction to the new `preview-spec-lessons` bucket (a required precondition, a route query param, a preview-auth step) — never a locator miss, which stays `aw-tester`'s `aw-tester-lessons`; the author reads **both** buckets so runner-discovered quirks make the next spec correct from the start; promotes to `diagnose`
```
mthines: suggestion (low): ⚪ **inventory miscounts the NOT RUN returns** **(non-blocking)**

`runner.md` has five `NOT RUN (…)` occurrences but only three distinct strings, which is what the table enumerates. So "four" is wrong either way, and "five returns" is four.

Evidence: `CLAUDE.md:68` (says four and five) · `README.md:106` (same wrong count)

<sup>`pr-reviewer` · commit `2b1df0a`</sup>



---

mthines (commented):
Reviewed `2b1df0a` — 4 findings, none blocking — 2 guard-integrity gaps in the new Step 2 checks, 2 half-done fixes. 4 finding(s) inline. Sticky exists but this access path cannot edit an issue comment (no `gh`, and no mcp issue-comment update tool) — report not updated in place; it is in the run log.

---

scripts/eval/l1.mjs
mthines: Confirmed both directions and fixed in `194076f`. This is the sharpest finding of the three passes: a guard added to close a vacuous guard was itself vacuous, in the same review it was introduced.

Measured the section directly rather than taking the numbers on trust:

```
Step2 start: 2260   next heading: 2513   section length: 253
window 400 => bleed into Step 3: 147 chars
offset of `terminal for this run` within section: 155
```

So my on-thread claim that these were "sliced from the heading, so a sentence elsewhere in the file cannot satisfy them" was wrong: the slice starts at the heading and never ends at the section. Took your end-bound shape, with `-1` handled on both `indexOf` calls (`RUN_NEXT === -1` falls back to `RUN.length`, so Step 2 being the last section is not a silent empty slice), and dropped both windows.

Also gave each check its own `-1` sentinel rather than leaving them to fail by side effect of a 1-char slice. You flagged that check 2 was "correct today by side effect" — that's the kind of correctness a later edit removes without anything naming what broke.

```
RED   V2/delegation-moved-to-step3       (want RED)
GREEN V3/benign-reflow                   (want GREEN)
RED   V1/heading-renamed                 (want RED)
```

---

scripts/eval/l1.mjs
mthines: Confirmed and fixed across `194076f` and `d63cf74`. All three are now guarded, section-bounded the same way as Step 2.

Worth recording that your U3 caught a fourth instance of the substring bug while I was fixing it. My first attempt asserted `/empty preview-spec block/` over the whole file — and that string lives on **two** surfaces in `review-loop`, the table row (664) and the Step 3 report slot (734), so deleting the row left L1 green with the check satisfied by the slot. Identical shape to the `"authored"` ⊂ `"not authored"` bug this PR already fixed once. It is now anchored to the row (`/^\|\s*`empty spec`[^\n]*\|[^\n]*empty preview-spec block/m`) with the slot arm asserted separately, since both surfaces are real requirements rather than one restated.

Full matrix, expectations declared up front so a green cannot be read as a pass after the fact:

```
RED   U3/empty-spec-row-removed          (want RED)
RED   U3b/slot-arm-removed               (want RED)
RED   U1/step1-gh-only                   (want RED)
RED   U2/catchall-removed                (want RED)
```

L1 1498 → 1504.

---

scripts/eval/l1.mjs
mthines: Confirmed and fixed in `194076f`. `grep -c` finds exactly one occurrence of the literal `inconclusive: …` in the file, so the assertion was pinned to the punctuation rather than the rule — in a block whose own comments argue against exactly that.

Took the spirit of your patch with the backtick boundary kept, so the match still has to land on a code-spanned outcome string rather than any prose use of the word:

```js
/`inconclusive:[^`]*`[\s\S]{0,80}terminal for this run/
```

`[^`]*` accepts `…`, `<reason>`, or any future spelling inside the span; the 80-char gap is now a within-sentence bound rather than a section window, which is what the 400 was doing badly.

```
GREEN V4/ellipsis-reworded               (want GREEN)
RED   V4b/terminality-sentence-deleted   (want RED)
```

The second is the control — wording-independence must not become rule-independence.

---

CLAUDE.md
mthines: Confirmed and fixed in `194076f`. Counted rather than re-guessed:

```
2 NOT RUN (chrome unavailable, user declined Playwright)
1 NOT RUN (no Chrome extension and no sub-agent dispatch available)
2 NOT RUN (sub-agent dispatch unavailable)
=> 5 occurrences, 3 distinct
```

You're right that "four" is wrong on either reading, and I introduced it — the number came from my own reply on the earlier thread ("all four `NOT RUN (…)` strings"), which I then carried into `CLAUDE.md` and `README.md` without counting. The table itself enumerates the correct three.

Both files now say "three distinct `NOT RUN (…)`", and `CLAUDE.md`'s "five returns that previously fell through" is corrected to four (`empty spec` + the three distinct strings).

---

scripts/eval/l1.mjs
```diff
@@ -5298,4 +5298,255 @@ const isPollBlock = (block) =>
...
+  s.check("G49 review-loop's remedy for it is an explicit URL, not waiting for a build",
+    /no deployment lookup on this access path[\s\S]{0,320}--url <preview-url>/.test(RL),
+    "`re-run once the preview is up` is unactionable here — nothing was looked up, so the"
+      + " human re-runs and gets the same string; only passing a URL changes the outcome");
+  // The other two rows added to that table landed as prose no check read, so reverting
+  // either left L1 green. An unmapped return is the mechanism that put a permanently-false
+  // note in a report once already, which is what makes the catch-all load-bearing rather
+  // than tidy: without it a delegate can gain an outcome and the table just drops it.
+  s.check("G49 review-loop's outcome table has a terminal catch-all row",
+    /\|\s*anything else\s*\|/.test(RL) && /unrecognised outcome/.test(RL),
```
mthines: issue (medium): 🟡 **catch-all's note arm is satisfied by the report slot** **(non-blocking)**

`unrecognised outcome` lives on line 669 and again in the Step 3 slot (734), so swapping the row's note for `treat it as a pass` keeps L1 green — the exact failure this check names.

Evidence: `skills/quality/review-loop/SKILL.md:669` (the row) · `skills/quality/review-loop/SKILL.md:734` (the slot, same string)

```js
-    /\|\s*anything else\s*\|/.test(RL) && /unrecognised outcome/.test(RL),
+    /^\|\s*anything else\s*\|[^\n]*unrecognised outcome/m.test(RL),
// then assert the slot arm separately, as the `empty spec` pair already is
// measured: replacing the note with `green` leaves L1 1504/1504 green
```

<sup>`pr-reviewer` · commit `d63cf74`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -5298,4 +5298,255 @@ const isPollBlock = (block) =>
...
+    RUN_I !== -1 && /preview-url-resolution\.md/.test(RUN_STEP2),
+    "the precondition is enforced in `preview-url-resolution.md`; a Step 2 that resolves the"
+      + " URL itself never reads that file, and the branch is unreachable however well it"
+      + " is written");
+  // Matches the RULE, not its punctuation. Requiring the literal `…` made a reword to
+  // `inconclusive: <reason>` turn L1 red while the rule was still honoured — the wording
+  // dependency this block's own comments disavow.
+  s.check("G49 runner.md Step 2 declares an inconclusive from that file terminal",
+    RUN_I !== -1
+      && /`inconclusive:[^`]*`[\s\S]{0,80}terminal for this run/.test(RUN_STEP2),
```
mthines: suggestion (low): ⚪ **the 80-char gap is the 400 at smaller scale** **(non-blocking)**

The gap uses 18 of 80, so one added qualifying clause (118 chars) turns L1 red while terminality still holds. Now that the slice is section-bounded, the second scope is what misfires.

Evidence: `skills/testing/preview-spec/rules/runner.md:33` (gap uses 18 of 80)

```js
-      && /`inconclusive:[^`]*`[\s\S]{0,80}terminal for this run/.test(RUN_STEP2),
+      && /`inconclusive:/.test(RUN_STEP2) && /terminal for this run/.test(RUN_STEP2),
// [^\n]* is the conservative middle, but it red-lights a 2-line split instead
// measured: reword green under both fixes, red under the 80; deletion still red
```

<sup>`pr-reviewer` · commit `d63cf74`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -5298,4 +5298,255 @@ const isPollBlock = (block) =>
...
+  const RUN = read("skills/testing/preview-spec/rules/runner.md");
+  // Bounded by the NEXT HEADING, never by a character count. A `slice(0, 400)` window was
+  // scoped in name only: Step 2 is ~253 chars, so 147 chars of Step 3 sat inside every
+  // assertion and deleting Step 2's delegation while naming the file in Step 3's opening
+  // kept L1 green — a vacuous pass in the very guard added to close a vacuous guard. The
+  // constant also failed the other way: `terminal for this run` sits ~155 chars in, so
+  // adding a couple of sentences of legitimate prose ahead of it pushed it out of the
+  // window and turned L1 red for a reflow that changed no rule.
+  const RUN_H2 = "## Step 2: Resolve the preview URL";
+  const RUN_I = RUN.indexOf(RUN_H2);
```
mthines: suggestion (low): ⚪ **both section slicers match the heading unanchored** **(non-blocking)**

`indexOf("## Step 2: …")` also matches inside `### Step 2: …`, so demoting the headings grows the slice to 7126 chars and both arms are satisfiable from Steps 4-5.

Evidence: `scripts/eval/l1.mjs:5430` (Step 2 slicer) · `scripts/eval/l1.mjs:5461` (Step 1, same hole)

```js
-  const RUN_I = RUN.indexOf(RUN_H2);
+  const RUN_I = (new RegExp(`^${escapeRe(RUN_H2)}\\s*$`, "m").exec(RUN) ?? { index: -1 }).index;
// same change in the Step 1 IIFE - one hole, two copies to fix
// length > 120 is a lower bound and cannot see over-capture
```

<sup>`pr-reviewer` · commit `d63cf74`</sup>



---

mthines (commented):
Reviewed `d63cf74` — 3 findings — 1 medium, 2 low; no blocking issues. 3 finding(s) inline. Sticky exists but this access path cannot edit an issue comment — report not updated in place.

---

scripts/eval/l1.mjs
mthines: Confirmed and fixed in `1d69ee8`. I proved it the harder direction before applying — not by deleting the row (which your own note shows the slot covers) but by **inverting** it:

```
=== invert the row, keep the slot ===
| anything else | treat as green — nothing to report |
GREEN -- rule inverted, guard silent (CONFIRMED)
```

That is the worse failure of the two: the guard stayed silent while the rule it names said the opposite of what it says. Deletion at least removes the row; inversion leaves a table that actively instructs the run to swallow an unmapped outcome.

Applied your patch and asserted the slot arm on its own surface, as the `empty spec` pair already does:

```js
/^\|\s*anything else\s*\|[^\n]*unrecognised outcome/m.test(RL)          // the row
/^Preview spec[^\n]*unrecognised outcome/m.test(RL)                    // the slot
```

Bite matrix, expectations declared before the run:

```
OK   M1 invert catch-all row              -> RED   (want RED)
OK   M2 delete catch-all row              -> RED   (want RED)
OK   M3 drop unrecognised arm from slot   -> RED   (want RED)
```

M1 was green before this commit.

Your structural point is the one I take most seriously and I've recorded it in the code: this is the **fifth** instance of one class in this block — a substring test whose satisfying occurrence is not the one the check names. The comment above the fix now says so, so the next author reads the class rather than the instance.

---

scripts/eval/l1.mjs
mthines: Confirmed and fixed in `1d69ee8`. Measured the gap rather than eyeballing it:

```
gap chars used: 18 of 80
```

So the window had 62 chars of headroom and no purpose — the slice is already section-bounded, which makes the character count a second, weaker scope that can only ever false-red. Dropped it entirely rather than widening it:

```js
/`inconclusive:[^`]*`[\s\S]*terminal for this run/.test(RUN_STEP2)
```

I went with `[\s\S]*` over your two-independent-tests form for one reason: ordering. Two `.test` calls accept a section that states terminality *before* the outcome code, which is not the rule — the rule is that the outcome named there is terminal. Within a section-bounded slice `[\s\S]*` costs nothing and keeps the sequence. Your `[^\n]*` middle is exactly right about the trade-off and is what ruled it out: it red-lights a two-line split, which is a reflow.

The control still holds — wording-independence must not become rule-independence:

```
OK   M7 reflow before terminality claim   -> GREEN (want GREEN)
```

and the terminality sentence's deletion stays red via the existing `V4b` case.

---

scripts/eval/l1.mjs
mthines: Confirmed and fixed in `1d69ee8`. Reproduced the over-capture directly:

```
=== does indexOf match a demoted heading? ===
matched at 2262   section length after demotion: 7126 (was 253)
```

So every assertion in the group ran against the whole document, and your parenthetical is the part that made it invisible: `length > 120` is a **lower** bound, structurally incapable of seeing an over-capture. It was checking for the failure mode that cannot happen here while the one that can went unmeasured.

Took your patch's anchoring and your "one hole, two copies to fix" note literally — extracted the shared slicer rather than fixing it twice:

```js
const runSection = (heading) => {
  const m = new RegExp(`^## ${heading.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}$`, "m").exec(RUN);
  if (m === null) return "";
  const next = RUN.indexOf("\n## ", m.index + 1);
  return RUN.slice(m.index, next === -1 ? RUN.length : next);
};
```

Both Step 1 and Step 2 now go through it. A level change yields an empty slice, which the existing sentinels catch — so the *lower* bound becomes load-bearing for the demotion case instead of blind to it, which is a nicer outcome than adding an upper bound (an upper bound would false-red on legitimate prose growth, the same mistake as the 400).

```
OK   M5 demote runner.md headings              -> RED   (want RED)
OK   M6 delegation deleted, named in Step 3    -> RED   (want RED)
OK   M8 Step 1 back to gh-only                 -> RED   (want RED)
```

M5 was green before this commit.

I also took the order-pinned check you reported but did not file — the `EMPTY`/`ABSENT` pair asserted the two arms of an unordered `<a | b | c>` enumeration in their written order, so swapping them turned L1 red for an edit that changed no rule. It is now two membership tests over the line that renders them, with a control confirming the swap is green (`M4b`).

---

scripts/eval/l1.mjs
```diff
@@ -5298,4 +5298,282 @@ const isPollBlock = (block) =>
...
+  // over-capture. Demoting every `## ` to `### ` in `runner.md` kept the match (at the
+  // `###` heading) and grew the "section" from 253 chars to 7126, so every assertion below
+  // ran against the whole document. The heading is therefore anchored to LINE START and to
+  // its exact level, and a level change now yields an empty slice, which the sentinels
+  // catch. One implementation, so the next defect is fixed once rather than twice.
+  const runSection = (heading) => {
+    const m = new RegExp(`^## ${heading.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}$`, "m")
+      .exec(RUN);
+    if (m === null) return "";
+    const next = RUN.indexOf("\n## ", m.index + 1);
```
mthines: issue (medium): 🟡 **`runSection` end bound ignores heading level and fences** **(non-blocking)**

The empty-slice argument covers this section's own heading only: promoting the next one from `## Step 3` to `# Step 3` grows the Step 2 slice 253 to 1813 chars with L1 still green at 1505/1505.

Evidence: `scripts/eval/l1.mjs:5442` (not level-aware) · `scripts/eval/lib.mjs:1` (stops correctly at 252)

```js
const RUN_STEP2 = extractSection(
  "skills/testing/preview-spec/rules/runner.md",
  "## Step 2: Resolve the preview URL",
);
```

<sup>`pr-reviewer` · commit `1d69ee8`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -5298,4 +5298,282 @@ const isPollBlock = (block) =>
...
+  // both copies carried the same latent defect: the heading was matched with `indexOf` on
+  // its bare text, which also matches INSIDE a demoted `### Step 2: …`, and the only bound
+  // on the result was a `length > N` LOWER bound — structurally incapable of seeing an
+  // over-capture. Demoting every `## ` to `### ` in `runner.md` kept the match (at the
+  // `###` heading) and grew the "section" from 253 chars to 7126, so every assertion below
+  // ran against the whole document. The heading is therefore anchored to LINE START and to
+  // its exact level, and a level change now yields an empty slice, which the sentinels
+  // catch. One implementation, so the next defect is fixed once rather than twice.
+  const runSection = (heading) => {
+    const m = new RegExp(`^## ${heading.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}$`, "m")
```
mthines: suggestion (low): ⚪ **Heading anchor is whitespace-exact** **(non-blocking)**

The accepted patch anchored with `\s*$`; the applied helper uses `$`, so an invisible trailing space on the heading empties the slice and reds three checks.

Evidence: `scripts/eval/l1.mjs:5439` (whitespace-exact anchor)

```js
new RegExp(`^## ${heading.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}\\s*$`, "m")
```

<sup>`pr-reviewer` · commit `1d69ee8`</sup>



---

mthines (commented):
Reviewed `1d69ee8` — 2 findings — the G49 anchoring fix works for the section's own heading but the new end bound still over-captures one heading over, and the description's L1 numbers are stale. 2 finding(s) inline. Sticky exists (issue comment 5559172827) but this access path has no issue-comment update tool — gh is not installed and the GitHub MCP grant exposes no comment edit — so the report was not updated in place.

---

scripts/eval/l1.mjs
mthines: Confirmed and fixed in `1882840`. You are right on both counts, and right that my resolving reply on `r3944070457` overclaimed: the empty-slice argument covers the section's **own** heading, which the anchored `exec` handles, and says nothing about the next one. Reproduced your numbers exactly:

```
M-I   promote `## Step 3` to `# Step 3`             GREEN   slice 253 -> 1813
M-I2  M-I + delegation sentence moved into that h1  GREEN   check satisfied from
                                                            outside Step 2 entirely
```

M-I2 is the one that matters — over-capture alone is latent, but with the sentence relocated the delegation check was being satisfied by a top-level section that is not Step 2, with L1 green. That is the vacuity, one heading over from where I had just closed it.

The sharper part of your finding is the one I had already written down and then failed to apply to the other end of the slice: `length > 120` is a **lower** bound and cannot see an over-capture. Two commits ago I retired the 400-char window for precisely that reason and then reintroduced the same blindness at the end bound.

Took your patch. The local slicer is gone rather than grown a third private copy of the scan:

```js
const runSection = (heading) => {
  try {
    return extractSection("skills/testing/preview-spec/rules/runner.md", `## ${heading}`);
  } catch { return ""; }
};
```

Two notes on how I applied it:

- **`extractSection`'s own start bound was the same bare `indexOf`** you flagged, so I fixed it *there* rather than working around it here — anchored to line start, but only when the `section` argument is a heading, since a non-heading section has no level to anchor to and two callers pass whole-file `null`. `G21g` already drives all seven L2 rubric headings through this helper and asserts a non-empty body, so the lib change arrives with real coverage rather than only my two new call sites.
- **The `try`/`catch` is deliberate**, not defensive noise: `extractSection` throws on a missing section while the sentinels below want an empty slice, so an uncaught throw would abort the whole L1 run before `s.report()` — the same failure mode `G21g` already guards against one thousand lines up.

Level-awareness proven by mechanism, not just by outcome:

```
under M-I, extractSection slice: 252 chars
last line: "Any `inconclusive: …` outcome there is terminal for this run"
```

Full matrix, expectations declared before the run:

```
OK   M-I2 M-I + delegation moved into the h1              -> RED    (want RED)
OK   M-J  fenced '## ' line before the terminality claim  -> GREEN  (want GREEN)
OK   C1   delegation sentence deleted                     -> RED    (want RED)
OK   C2   terminality sentence removed                     -> RED    (want RED)
OK   C3   heading renamed                                  -> RED    (want RED)
OK   C4   Step 2 demoted to h3                             -> RED    (want RED)
```

One correction to my own expectations: I declared `M-I` alone as `want RED` and it came back GREEN — my error, not a residue. Promoting a heading moves no content, so nothing about the rule changed and GREEN is correct; it was green *by over-capture* before and is green *by correctness* now, which is why the slice length above is the evidence and the pass/fail is not.

I also took your deferred nitpick on the fence case (same root cause, same fix) — `extractSection` skips fenced blocks, so a ```text block inside Step 2 containing a `## ` line no longer truncates the slice.

---

scripts/eval/l1.mjs
mthines: Confirmed and fixed in `1882840`. You are right that I dropped the `\s*` from the patch I said I was taking:

```
M-A   trailing space on the Step 2 heading   RED (3 checks)  ->  GREEN
```

An invisible, meaning-preserving edit reddening three checks is squarely the "can only false-red" class this block's comments disavow, and it is the third instance of it in this PR after the 400-char window and the 80-char gap.

The fix lives in `extractSection` with the rest of the start bound, tolerating trailing spaces and tabs specifically rather than `\s*`:

```js
new RegExp(`^${escaped}[ \\t]*$`, "m")
```

`\s` in a non-`s` regex still matches `\n`, so `\s*$` with the `m` flag would let the anchor slide across a blank line and match a heading's *prefix* against a following line's content. `[ \t]*` is the invisible-whitespace class and nothing else, which is exactly the meaning-preserving set. Renaming or demoting the heading still reds (`C3`, `C4` above), so tolerance did not become indifference.

I also took your other deferred nitpick, `M-H`: the catch-all row check was pinned to the cell's exact text, so bolding it to `| **anything else** |` reddened a row that says the same thing. It now matches the cell's *content* within the row's first column (`^\|[^\n|]*anything else[^\n|]*\|`) — anchoring is about which row satisfies the check, and the decoration carries no meaning. Inverting the row's note still reds (`C5`).

---

mthines:
Added the durable fix the review proposed and then declined as below its materiality bar — the lint over the `G49` block. `c76d6d1`.

**Its own logic found a live defect before it existed.** Enumerating the block to remove every exemption ahead of the lint turned up three multi-surface bare assertions; bite-testing two of them surfaced an **eighth** instance of the class this PR is about:

```
/pass --url/.test(PS)   →  2 hits in preview-spec/SKILL.md, lines 105 and 135
```

Line 105 is Step 0's outcome string — the surface the check names. Line 135 is Operation `run`'s `--url` prose. Deleting it from line 105 and keeping 135 left **L1 green**. (The other candidate, `/no deployment lookup on this access path/.test(RL)`, was *not* vacuous: the `{0,320}` gap held. It has since been row-anchored anyway, so the last character-gap-as-scope is gone.)

**What the lint asserts.** Every `.test()` in the `G49` block against a whole-file `read()` binding must be line-anchored — `^` in the body, `m` in the flags — which pins the match to one line or one table row. Bindings are **derived** from the block's own `read()` calls, so one added tomorrow is covered without editing the lint. It is fail-closed: a receiver it cannot parse as a regex literal is a violation, not an exemption, because unparseable is the hole a hoisted pattern constant or a `new RegExp(...)` would slip through.

Three self-sentinels stop the lint from passing vacuously itself (block located and > 10k chars, ≥ 3 bindings derived, ≥ 10 `.test()` calls seen). Deliberately **no** floor on the whole-file count: that is the quantity the rule drives to zero, and a floor would redden on exactly that improvement.

**Made the block conform rather than exempting it**, so the lint arrives with no allowlist. Step 0's two access-path checks, the access-path precondition pair, and `create-pr`'s Step 6.4 / Step 10 checks now read section slices via a local `section()` wrapper over the shared `extractSection` (degrading to `""` on a renamed heading, with a sentinel proving each slice non-empty); the two surviving `RL` checks are row-anchored.

**Bite-tested, five mutations, every expectation declared before the run, all wanting RED and all RED:**

| Mutation | Result |
|---|---|
| a re-pointed check reverted to a bare `.test(PS)` | RED — `l1.mjs:5421 .test(PS) is line-anchored` |
| `^` and `m` dropped from a row-anchored `.test(RL)` | RED — `l1.mjs:5540 .test(RL) is line-anchored` |
| receiver replaced with `new RegExp(…, "m")` | RED — `tests a whole file through a regex literal` |
| the `G49` banner renamed | RED — all three sentinels, `banner=-1 len=0` |
| the binding-derivation regex broken | RED — `derived=[]` |

The last two also drop the check total by 5, since the per-assertion checks vanish when the scan finds nothing — which is the point of pairing them with named sentinels.

L1 **1474 → 1517**, `l1` and `l2` green on `c76d6d1`. PR description updated: the "Not in this diff" note about the lint is gone, replaced by a `5.` in *What changed*, the eighth row in the vacuity table, and a new stated limitation — **the lint is scoped to the `G49` block alone**, located by its exact banner prefix (`// ── G49: `, so it cannot pick itself up). Other blocks in `l1.mjs` still assert against whole files unlinted; generalising the rule is a separate change with its own baseline to establish.

Still draft. Undrafting remains a human decision.

---

mthines:
**`l2` is red on `c76d6d1`, and it is not this PR's failure.** Standing down on it, with evidence rather than a re-run.

```
code-review-retrieval-relevance: 3/5 (60.0%)
✗ a suite is below the EVAL_GATE floor of 70%
```

Everything else in that run is green or comfortably above the floor. `l1` is **green** on the same commit (1517/1517, [run 594](https://github.com/mthines/agent-skills/actions/runs/34040976511)).

**Why it is not this PR's.** Dispatched `evals-l2.yml` on `main` at `b6fee49` — the base — and it fails identically:

| ref | `code-review-retrieval-relevance` | run |
|---|---|---|
| `c76d6d1` (this PR's head) | 3/5 (60.0%) | [34040976517](https://github.com/mthines/agent-skills/actions/runs/34040976517) |
| `b6fee49` (`main`) | 3/5 (60.0%) | [34041417796](https://github.com/mthines/agent-skills/actions/runs/34041417796) |

Same suite, same count, same two misses. This PR's diff is `scripts/eval/l1.mjs` only; that suite's inputs are `agents/pr-reviewer/rules/memory.md` (read live) and `scripts/eval/golden/code-review-retrieval-relevance.jsonl`, neither of which the diff touches.

**Why it appeared now rather than four commits ago.** It is a first execution, not a regression. The `l2` job's eval step ran for **0 seconds** on `1882840` and **6 minutes** on `c76d6d1` — the 0-second path is `l2.mjs`'s clean `no ANTHROPIC_API_KEY — skipping` exit, which the workflow deliberately treats as a pass so the check is safe to require and safe on fork PRs. The secret became available in CI between 13:55 and 15:01, so this is the first run in which any L2 suite actually executed. Every "l2 green" on this PR before now was a skip. That is worth knowing on its own: a check that passes by skipping is indistinguishable from a check that passes, which is the same `F6` shape this PR is about — one layer up, in the harness rather than the guard.

**No fix exists to port, and I am not widening the PR to write one.** The two misses are both `expected: surface, got: skip`, and both turn on one rubric claim: Step 1.0's `memory_list` read is **list-reachable** — not `seen_count`-gated and not diff-filtered — so a lesson under one of the two read tags surfaces regardless.

| case | candidate | model said |
|---|---|---|
| `reviewer-lesson-list-reachable-unrelated-diff` | `loop::reviewer-lessons`, `seen_count: 5`, diff is unrelated E2E scaffolding | `skip` — filtered on diff relevance |
| `reviewer-lesson-seen2-below-threshold` | `loop::reviewer-lessons`, `seen_count: 2` | `skip` — filtered on the promotion threshold |

Both golden `notes` state the distinction explicitly (`seen_count` marks promotion-grade should-fire, it is not a surfacing filter; diff-relevance governs only the *additional* Step 1.2c enriched-search path, an OR, never a filter that can remove a list-reachable lesson). So the rubric text is not conveying a distinction its own golden set says it makes — a real finding, and a real fix would be either a rubric edit or a golden-set correction.

Which of those it is, is a judgment call about a gate, so it does not get made silently inside this diff. Two further facts a fix should weigh, neither of them something to decide in a CI-fix commit:

- The set has **5 records**, and the workflow's own comment justifies the 70% floor on the assumption it "only trips on a badly-broken rubric, not 1–2 cases of model noise on these < 50-case sets". On a 5-record set, 2 misses is 40% — the assumption does not hold at this size, and `scripts/eval/README.md` already flags sets under 50 as report-only-ish.
- Its `notes` say **BOOTSTRAP SEED — NOT A REAL BASELINE** on every record.

The right home for it is a `/eval-iterate` pass on its own branch: classify each miss `code-bug` vs `eval-bug`, fix the minimal cause, and require two consecutive passing re-runs. `mergeable_state` is `unstable`, not `blocked`, so `l2` is not gating this PR's merge.

Still draft. Undrafting remains a human decision.

---

mthines:
Two corrections to my comments above, then the merge plan.

**Correction 1 — `l2` was never green on `c76d6d1`.** [The `c76d6d1` comment](https://github.com/mthines/agent-skills/pull/181#issuecomment-5560088267) says "`l1` and `l2` green on `c76d6d1`". `l1` was; `l2` was not — I read a `success` conclusion that was in fact the 0-second `no ANTHROPIC_API_KEY — skipping` exit. [The next comment](https://github.com/mthines/agent-skills/pull/181#issuecomment-5560168694) established that within the hour, but the false line stood. The PR body's copy of it is now fixed.

**Correction 2 — the `l2` work is no longer on a separate branch.** That comment said the right home was "a `/eval-iterate` pass on its own branch". Three commits of it landed here instead, at the author's direction:

| commit | what |
|---|---|
| `852c158` | re-point the retrieval-relevance rubric at the two subsections its `instruction` names; add `rubric.sections` + fail-closed `rubricFor()`; `G21g` floor → parity check; `--suite <typo>` exits 1; workflow `suite` dispatch input |
| `be03c94` | let `G21g`'s `section:` parse tolerate a trailing comma (a latent quiet exemption the retired floor would have hidden) |
| `d72d196` | state in `### 1.0` that the read filters on tag/scope/expiry only — no diff parameter, no `seen_count` |

**Both hypotheses were falsified.** [34044758211](https://github.com/mthines/agent-skills/actions/runs/34044758211) and [34044924487](https://github.com/mthines/agent-skills/actions/runs/34044924487) each returned an identical `3/5 (60.0%)` with the same two misses. The three commits stand on their own merits — a suite fed a rubric 2.5× broader than its own question, prose that never stated its read's non-filters, and three ways for the harness to grade nothing while reporting green — but none of them is the gate's cause. `/eval-iterate` stopped at 2 of 5 iterations rather than spending a third guess: `escalated`, not `confirmed-green`.

The other eight suites are unaffected — all clear the floor on [34044749700](https://github.com/mthines/agent-skills/actions/runs/34044749700), lowest `shape-depth-routing` at 81.8%.

**Merging with `l2` red, deliberately.** It is red identically on `main`, this diff does not touch the suite's golden set, and `mergeable_state` is `unstable` rather than `blocked`. The remaining half — the `instruction` rewrite **plus** genuine `skip` decoys, which have to ship together or the 4-`surface`/1-`skip` split turns the rewrite into an answer key — goes to a follow-up PR. No golden case was edited, skipped or added here, and `EVAL_GATE` is untouched.

---

PR #182, opened 2026-09-06
Make the evals affected-only, opt-in, and observable — then run them for the first time

mthines:
The evals existed but had never measured anything. `ANTHROPIC_API_KEY` was unset, so 232 consecutive L2 runs were green having made zero model calls, and `l2-detection.mjs` ran in no workflow at all — only its `--self-test` was ever exercised. This makes the suites cheap enough to want to run, opt-in so they cost nothing unasked, observable in Dash0, and then actually runs them.

## What changed

**Selection is derived, never mirrored.** `select-suites.mjs` maps a PR's changed files to suite names off each suite's own `rubric.file` / `golden` in `suites.mjs`, and CI runs one job per affected suite. The workflow's hand-written `paths:` filter is gone — it had drifted to four of nine rubric files missing, so those four could be edited with their own eval never running. A harness file selects *all* suites: the deliberate fail-open direction, since extra suites cost tokens while a missed suite costs coverage and reports green either way.

**L2 is opt-in.** Suites run only on the `run-evals` label or a `workflow_dispatch`. The `pull_request` trigger deliberately stays: the trigger governs whether the *check appears*, the gate governs whether *suites run*, and a dispatch-only workflow never reports on a PR — so requiring its check would leave every PR pending forever. An unlabelled PR now pays one ~3s bookkeeping job with no checkout and no API call.

**A run someone asked for cannot report green having measured nothing.** The opt-in path sets `EVAL_REQUIRE_KEY=1` and exits 3 on a missing key. A fork or fresh clone still skips cleanly. Two sibling doors are closed the same way: an unknown `--suite` exits 2, and a suite selected with a missing or empty golden file exits 4 rather than printing `(skip …)` or scoring `NaN%`.

**`bug-detection` is wired in, in its own job, hard-gated** on `recall ≥ 0.7` / `fp ≤ 0.2` (unlike the suites' soft floor — those two rates *are* its contract). It rides the same derivation without becoming a matrix entry, via a `DETECTION` declaration in `suites.mjs`.

**Telemetry.** L2 runs emit OTLP traces and metrics (`service.namespace=agent-skills`, `service.name=evals`), because stdout answers *what was today's accuracy* and cannot answer *has this rubric been drifting for three weeks*. Off unless `OTEL_EXPORTER_OTLP_ENDPOINT` is set; a miss is the measurement, so a wrong answer is never a span error; the flush precedes the gate exit, so a run that trips the gate still ships the trace you most want to read. Each export request carries a 10 s timeout — Node's `fetch` has none, and `flush()` is awaited *before* that gate exit, so an unresponsive ingress would otherwise hold the run open until the job cap killed it and take the accuracy verdict with it. The detection job is deliberately **not** given these variables: its runner has no exporter, so an endpoint and a bearer token there would be inert — and one of them is a credential placed in an environment nothing reads.

**The scorer had two defects, and each invalidated a different thing.** A reply enumerating every choice now reads as ambiguous rather than as the first one — the old earliest-substring rule scored the `MODE SELECTION:` template line's own `[Micro | Lite | Full]` as a confident `Micro`, which is how five `tier-routing` misses landed on one label. And a reply naming a choice that *contains* another (`optimal` inside `suboptimal`, `promoted` inside `not-promoted`) resolved to neither: `named.length === 1` could never hold, so in the two nested-choice suites every reply but the byte-exact one scored as a miss. The parse now prefers the longest match. Misses print the model's **raw reply**, since a line showing only the parsed label cannot distinguish a wrong answer from a wrong parse. Rubrics travel as a `cache_control: ephemeral` block for ~80% off the input bill. Suites under 10 cases are labelled `[advisory]` and cannot breach the floor — at 5 cases a 70% gate allows exactly one miss.

## Measured results — the first full CI run

The label fired all nine suites plus `bug-detection`. **131/140 = 93.6%**, against a 130/140 baseline. Every suite:

| suite | cases | baseline | CI run | misses |
| --- | --- | --- | --- | --- |
| `tier-routing` | 30 | 25/30 | **30/30** (100%) | — |
| `bug-class` | 18 | 18/18 | **18/18** (100%) | — |
| `aw-should-trigger` | 15 | 15/15 | **15/15** (100%) | — |
| `reviewer-agreement-bump` | 6 | 6/6 | **6/6** (100%) `[advisory]` | — |
| `optimize-approach-optimality` | 15 | 15/15 | **14/15** (93.3%) | `oa-premature-generic` |
| `shape-depth-routing` | 22 | 20/22 | **20/22** (90.9%) | `mutex-small-delta`, `no-deep-pass-on-record` |
| `severity-tiering` | 15 | 15/15 | **13/15** (86.7%) | `sev-5`, `sev-15` |
| `complexity-triage` | 14 | 13/14 | **12/14** (85.7%) | `simple-3`, `simple-5` |
| `code-review-retrieval-relevance` | 5 | 3/5 | **3/5** (60%) `[advisory]` | `…unrelated-diff`, `…seen2-below-threshold` |

**The near-flat aggregate is the finding, and it is a caution about reading aggregates.** `tier-routing` gained five cases from a rubric fix; four suites nothing in this PR touched each drifted −1 or −2. The total moved by one. A reader watching only the total would have seen a flat line across the most substantive rubric fix in this repo's history — so the composition is what to read, and only `tier-routing`'s +5 is attributable to a change here.

**Two of the three re-measured suites came back a case worse in CI than in the local post-fix run** — `shape-depth-routing` 20/22 not 21/22, `code-review-retrieval-relevance` 3/5 not 4/5 — with nothing changed but the sample. That is the ±1–2 case variance this PR's own notes warn about, arriving immediately and on two suites at once. The docs now label the local post-fix figures **directional** and the CI columns authoritative. `tier-routing`'s 30/30 reproduced exactly and is the one result strong enough to quote on its own; it is also the largest suite.

**A second CI run reproduced all three figures exactly** — including `shape-depth-routing` re-missing the *same two* cases rather than two of comparable count. That distinction is worth the column it costs: a total that repeats with an identical miss set is a rubric defect, where one that moves around at the same total is sampling noise. So `mutex-small-delta` and `no-deep-pass-on-record` are real gaps for whoever next works that rubric — and the second of them still routing `deep`→`quick` means `D6` is being missed even now that it is its own numbered row, which is a useful negative result about the checklist fix.

`tier-routing`'s five misses were all *under-tierings* and all five closed. Nothing was re-labelled and no golden case was added — the labels were right and the rubric was wrong, which is the outcome that suite exists to distinguish from model noise.

**`shape-depth-routing` took two edits, and the intermediate run is the finding.** The first fixed the refresh clause's grammar; `refresh-runs` duly passed and the suite came back **one case worse**, with two different misses. One miss becoming two identified the defect as structural: both were the triggers buried deepest in a table cell holding thirteen `·`-separated conditions, and clarifying a clause *inside* that cell pushed everything after it further out of reach. The conditions are now an enumerated `D1`–`D13` checklist placed *before* the tier table — same triggers, same order, same thresholds. A single-suite iteration that looks like a fix can be a net regression, and the count is what says so.

**The cache warning was the defect, not the cache.** Three suites reported `cache MISSED` — and `README.md` had already predicted by name which three cannot benefit, all under the ~1024-token minimum cacheable prefix. The run confirmed exactly those three and no others; the other six each wrote and read a 1,179–8,185 token block. But the message read as a defect report — *"the rubric is being re-billed per case"* — for a case where there is nothing to discount and nothing to fix. The two are now separated: `cache not applicable`, versus `cache MISSED` kept for a long prompt with no cache activity, which is the one worth investigating.

The reviewer then caught that the *arithmetic* behind that split was still wrong — it decided on total input ÷ case count, but only the system block carries `cache_control`, so a small rubric with large cases could be dragged over the bound and cry `MISSED` anyway. The runner now measures the largest system block it actually built, and the next CI run printed `cache not applicable (largest cached prefix ~935 tokens, …)` for `tier-routing`, matching the local measurement to the token. That also corrects a claim in the code: three suites fall under the bound and they are **not** all cheap — measured, they are ~935 / ~405 / ~317, so `tier-routing` misses by under 90 tokens while carrying the largest uncached bill in the run (30,168 input tokens, more than any cached suite pays *after* its discount). Padding a rubric to reach the bound would be writing for the biller instead of the reader, so the loss is named and accepted rather than engineered away.

## The detection gate is red, and I left it red

| stage | run 1 | run 2 (CI) | run 3 (CI) |
| --- | --- | --- | --- |
| finder only | 75% (15/20) · fp 50% | 80% (16/20) · fp 50% | 70% (14/20) · fp 40% |
| + verifier | 75% · fp **30%** | 60% (12/20) · fp **40%** | 70% · fp **30%** |
| verifier lift | **+20** (fp −20, recall −0) | **−10** (fp −10, recall −20) | **+10** (fp −10, recall −0) |

Red in all three — on `fp_rate` alone in runs 1 and 3, on both rates plus a negative verifier lift in run 2. **This describes the core, not this change.** `finders.md` and `finding-verifier.md` are untouched here; this PR's only edit to `l2-detection.mjs` is the `EVAL_REQUIRE_KEY` guard. I did not lower the gate and did not edit the two rubrics: that is a rubric change needing its own golden decoys and its own re-run per the maintenance table. Lowering the gate is the fix-to-pass forbidden everywhere else here, and is what left the eval unrun to begin with.

**Run 2 retracts run 1's headline claim, and that retraction is the most useful thing the runs produced.** Run 1 showed the verifier halving the finder's false positives at *zero* recall cost, which read as `finders.md`'s "finders flag, the verifier filters" polarity measured rather than argued — and it was written down that way, in both `README.md` and `CLAUDE.md`. Run 2 shows it spending 20 points of recall to buy 10 of precision. Run 3 comes back looking like run 1. So two of three runs show the polarity working and one shows it inverted, across a lift spread of −10 to +20 points — which at 20 seeded records (one case = 5 recall points) and 10 controls (one control = 10 fp points) is the entire spread. **The verifier's value is still not established**, though run 2's recall cliff now reads as the outlier rather than the rule. Both docs are corrected: no single detection run may be quoted for or against the verifier, and whoever raises the core's precision must run repeats before *and* after — three is a floor, since it took the third run to tell run 2's cliff from a trend.

**What the three runs do establish is where the work goes.** `fp_rate` after the verifier has never once reached the 20% gate (30 / 40 / 30) while recall straddles its 70% bar (75 / 60 / 70). The precision gap is the reproducible failure and the recall gap is not yet distinguishable from noise, so the controls are the target. `intent-mismatch` is the weak class in all three (1/4, 2/4, 2/4 against 3/4–4/4 elsewhere) with most misses never flagged, so that gap is finder recall rather than verifier over-filtering; `standards` is 4/4 every time. And one record — `intent-revert-not-fix` — failed on output *shape* (prose instead of JSON) in runs 1 and 3 both, counted as a miss and never as clean. That it recurs on one specific record points at that record's prompt rather than model flakiness, which is cheap to check before anyone touches the parser.

## Guards

`CLAUDE.md` gained a mandatory maintenance table: a change altering a decision the evals measure must update the evals in the *same* commit, with an exhaustive lookup of what you touched → what you owe. Opt-in created two new obligations, both stated — nobody else will notice a rubric you did not verify, and `select-suites.mjs` on your own changed files tells you which suites you owe.

New L1 checks (**1587 total, green**): `G21m` guards the detection wiring end to end, `G21i`/`G21l` the opt-in and the scorer, `G21k` the telemetry rules, `G21d`/`G21h` the derivation and the absence of a rubric-path mirror.

Each new check was **probed by breaking the thing it guards**. Two did not bite first time. `G21m`'s aggregator check originally tested for `exit 1` anywhere in the `l2` job and stayed green when I downgraded the detection failure branch to an `echo`, because the gate and selection branches already carry one; it now asserts on the failure path specifically. And `G21l` `eval()`s the live `parseChoice` but exercised it only with `["Micro","Lite","Full"]`, where no choice contains another — so the containment defect above passed it green. It now probes the nesting in both directions and both live shapes (a substring inside a word, a hyphenated prefix), plus the documented residue: for a nested pair, containment is the only evidence available, so an unbracketed enumeration resolves to the longer choice rather than reading as ambiguous. That is a real trade, taken deliberately and pinned so it cannot be "fixed" back.

The link guard also earned its place mid-review: renaming the README's detection-baseline heading broke a `CLAUDE.md` anchor and L1 went 1586/1588 immediately, naming the file and line.

## Verification

The CI wiring was exercised for real in **all three** directions, which matters because the aggregator is `if: always()` — a failed dependency arrives as a value to *read*, so `needs:` alone does not propagate failure. Run 245: detection failed, `l2` failed with it. Run 247: detection `skipped`, `l2` passed. And both opted-in runs: nine suites green, detection red, **`l2` red** — the load-bearing case, on a real failure rather than a forced one, now twice.

`l2` remains the one check to require — never a per-suite job, which comes and goes with the diff and would leave a required check pending forever.

**Eval obligation — discharged, twice.** The nine suites and `bug-detection` ran on the opt-in path, and the review round that followed touched two more measured surfaces: `l2.mjs` (a harness file, which per this PR's own table owes all nine plus detection) and `depth-routing.md` (`shape-depth-routing`'s live rubric). Both changes land where they cannot move a label — a reporting-path fix after every model call, and two inline definitions in the rubric's prose — but the rule is deliberately mechanical rather than judgment-based, so the push re-ran the full set rather than arguing the exception. It came back 9/9 suites green with all three quoted figures reproduced exactly, and detection red at 70% / 30%. Both numbers are in `README.md`.

Baselines are recorded in `scripts/eval/README.md` — compare against those rather than a remembered number.

---
https://claude.ai/code/session_01EFoqtKWwjJcRFDxQ1Aowss


mthines:
### 🟠 6 findings — 1 blocking

Makes the L2 evals affected-only, opt-in and OTLP-observable, then runs them for the first time — but the rewritten answer scorer mis-parses the two suites whose choice words nest.

**Blocking:** 1 blocking finding — the new scorer scores a correct reply as unparseable on `optimize-approach-optimality` (15 cases, gating) and `reviewer-agreement-bump`

|---|---|---|
| Nested choices score every non-bare correct reply a miss | `scripts/eval/l2.mjs:116` | 🟠 high · blocking |
| G21l cannot bite: it probes only non-nested choices | `scripts/eval/l1.mjs:1592` | 🟡 medium |
| OTLP export has no timeout | `scripts/eval/telemetry.mjs:241` | 🟡 medium |
| No maintenance row for a harness-file change | `CLAUDE.md:340` | 🟡 medium |
| `hrNs` is dead code | `scripts/eval/telemetry.mjs:53` | ⚪ low |
| `@returns` omits `detection` | `scripts/eval/select-suites.mjs:35` | ⚪ low |

<details>

**Needs attention**

| Gate | Status | Details |
|---|---|---|
| Description vs. code | ✅ | The description matches the diff, including the red detection gate it declines to lower. |
| Documentation | ✅ | README baselines, a CLAUDE.md maintenance table, and rationale comments at each non-obvious decision. |
| Self-review signals | ✅ | No debug logs, leftover markers, conflict residue, or unjustified suppressions on new lines. |
| Code review | ❌ | 1 blocking finding at `scripts/eval/l2.mjs:116`, plus 5 non-blocking. |

**Found**

Quality — produced 8 → posted inline 6 · cleared 6 · carried forward 0 · deferred 0 · below-bar 0
Severity — 🟠 1 high · 🟡 3 medium · ⚪ 2 low
Standards — ran · 2 governing doc(s) · 1 finding

**Run**

full · 2332 lines in delta · tier deep · depth checkout · blast_radius=none, semver_delta=none, high_stakes=0, risky_shapes=none — first review of this PR
Integrations — not activated (no manifest or lockfile in the diff)
CI — No check runs have reported on this head, and GitHub reports the branch as conflicting with `main` (`mergeable_state: dirty`). Reported, not graded.
Memories — 0 indexed · 0 used

<sup>Nothing to report — optimality (5 judged), measurability (skipped), 0 files skipped.</sup>

</details>

<sup>`pr-reviewer` · commit `83fc3a7` · full review · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md) · updated <relative-time datetime="2026-09-06T17:40:00.000Z">Sep 6, 2026 5:40pm UTC</relative-time></sup>

---

scripts/eval/l2.mjs
```diff
@@ -10,130 +10,126 @@
...
-  }
-  return best || `?(${text.slice(0, 24)})`;
+  const t = text.trim();
+  const eq = choices.find((c) => c.toLowerCase() === t.toLowerCase());
+  if (eq) return eq;
+  // A bracketed placeholder is scaffolding, not a claim: `Tier: Full [not Micro]`
+  // names one choice, and `[Micro | Lite | Full]` names none.
+  const low = t.replace(/\[[^\]]*\]/g, " ").toLowerCase();
+  const named = choices.filter((c) => low.includes(c.toLowerCase()));
+  if (named.length === 1) return named[0];
```
mthines: issue (high): 🟠 **Nested choices score every non-bare correct reply a miss** **(blocking)**

`optimal` is a substring of `suboptimal` and `promoted` of `not-promoted`, so `named.length === 1` never holds once the exact-match branch is missed. A reply of `suboptimal.` returns `?(…)`.

Evidence: `scripts/eval/suites.mjs:61` (optimize-approach-optimality choices) · `scripts/eval/suites.mjs:69` (reviewer-agreement-bump choices)

```js
const named = choices.filter((c) => low.includes(c.toLowerCase()));
// prefer the longest match: drop any choice contained in another matched choice
const top = named.filter((c) => !named.some((o) => o !== c && o.toLowerCase().includes(c.toLowerCase())));
if (top.length === 1) return top[0];
```

<sup>`pr-reviewer` · commit `83fc3a7`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -1389,21 +1400,325 @@ function checksInSync(plan, checks) {
...
+    // Every check here is executed against the runner's real behaviour, not its prose.
+    {
+      // The parse is pulled out of the live source and run, so a rewrite that
+      // reintroduces earliest-wins fails here rather than in a quarterly review.
+      const fnSrc = l2runner.slice(l2runner.indexOf("function parseChoice"), l2runner.indexOf("const summary = []"));
+      let parseChoice = null;
+      try { parseChoice = eval(`(${fnSrc.slice(fnSrc.indexOf("function parseChoice"))})`); } catch { /* reported below */ }
+      s.check("G21l parseChoice is extractable and callable from the live runner", typeof parseChoice === "function");
+      if (typeof parseChoice === "function") {
+        const C = ["Micro", "Lite", "Full"];
```
mthines, line 1609: issue (medium): 🟡 **G21l cannot bite: it probes only non-nested choices** **(non-blocking)**

The check `eval()`s the live `parseChoice` but exercises it solely with `["Micro","Lite","Full"]`, where no choice contains another — so a containment defect passes it green.

Evidence: `scripts/eval/l2.mjs:116` (the rule this check guards)

```js
const N = ["optimal", "suboptimal"];
s.check("G21l a choice that CONTAINS another still parses to the one that was said",
  parseChoice("suboptimal.", N) === "suboptimal");
```

<sup>`pr-reviewer` · commit `83fc3a7`</sup>



---

scripts/eval/telemetry.mjs
```diff
@@ -0,0 +1,370 @@
...
+          }],
+        },
+      });
+    }
+    if (metrics.length === 0) return null;
+    return { resourceMetrics: [{ resource: { attributes: this.resource }, scopeMetrics: [{ scope: { name: "agent-skills/evals", version: "1" }, metrics }] }] };
+  }
+
+  async #post(path, body) {
+    const res = await fetch(`${this.endpoint}${path}`, { method: "POST", headers: this.headers, body: JSON.stringify(body) });
```
mthines: suggestion (medium): 🟡 **OTLP export has no timeout** **(non-blocking)**

Node's `fetch` never times out on its own, so an unresponsive ingress stalls `flush()` — awaited before the gate exit — until the 20-minute job cap kills the run.

Evidence: `scripts/eval/l2.mjs:249` (awaited before gate exit)

```js
const res = await fetch(`${this.endpoint}${path}`, {
  method: "POST", headers: this.headers, body: JSON.stringify(body),
  signal: AbortSignal.timeout(10_000),
});
```

<sup>`pr-reviewer` · commit `83fc3a7`</sup>



---

CLAUDE.md
```diff
@@ -322,11 +322,38 @@ All applicable hops must resolve. If any is missing, the harness will not see th
...
+
+**A change that alters a decision the evals measure MUST update the evals in the SAME commit.**
+An eval is not follow-up work: a rubric edited without its suite re-run is an unverified behavioural change, and a suite left pointing at a deleted rubric is a red gate on the next author's PR.
+
+"Big change" is not the trigger — these are, and they are exhaustive.
+Look up what you touched:
+
+| You changed… | You MUST… |
+| --- | --- |
+| the body of a section a suite reads (any `SUITES[].rubric.file` + `section`) | run that suite (`node scripts/eval/l2.mjs --suite <name>`) and report the accuracy. Re-label or add golden cases if the decision boundary moved; a rubric edit that changes no label is the normal case, and saying so is the deliverable |
```
mthines, line 340: suggestion (medium): 🟡 **No maintenance row for a harness-file change** **(non-blocking)**

The rows are declared exhaustive, but none covers `l2.mjs` / `suites.mjs` / `select-suites.mjs` — where the selector's own rule is that one such change selects all nine suites.

Evidence: `scripts/eval/select-suites.mjs:45` (harness selects every suite)

```markdown
| a HARNESS file (`SUITES.HARNESS_FILES`) | run **every** suite — the selector's rule 1. This change re-ran three of nine |
```

<sup>`pr-reviewer` · commit `83fc3a7`</sup>



---

scripts/eval/telemetry.mjs
```diff
@@ -0,0 +1,370 @@
...
+//
+// FAILURE POLICY: every export error is swallowed and reported to stderr only. An
+// eval must never go red because a telemetry backend was unreachable — the accuracy
+// number is the product, the span is the receipt.
+import { randomBytes } from "node:crypto";
+import { hostname } from "node:os";
+
+const HEX = (bytes) => randomBytes(bytes).toString("hex");
+const nowNs = () => String(BigInt(Date.now()) * 1_000_000n);
+const hrNs = (ms) => String(BigInt(Math.round(ms * 1e6)));
```
mthines: nitpick (low): `hrNs` is defined here and never referenced anywhere in the module. **(non-blocking)**

<sup>`pr-reviewer` · commit `83fc3a7`</sup>

---

scripts/eval/select-suites.mjs
```diff
@@ -0,0 +1,187 @@
...
+// suites.mjs (SUITES.DETECTION) and reported as a separate boolean, because it is a
+// separate runner with its own gates rather than a matrix entry.
+import { SUITES, HARNESS_FILES, goldenPath, detectionInputs } from "./suites.mjs";
+
+const ALL = SUITES.map((s) => s.name);
+const DETECTION_INPUTS = detectionInputs();
+
+/**
+ * @param {string[]} changed  repo-relative changed paths
+ * @returns {{ suites: string[], reason: string, matches: Record<string, string[]> }}
```
mthines: nitpick (low): The `@returns` type omits `detection`, which `selectSuites` returns and the workflow reads. **(non-blocking)**

<sup>`pr-reviewer` · commit `83fc3a7`</sup>

---

mthines (commented):
---

scripts/eval/l2.mjs
mthines: Addressed in `4dfc9bc`. Confirmed by executing the extracted function against both nested pairs — `suboptimal.` → `suboptimal`, `not-promoted.` → `not-promoted`, and the contained choice stays reachable (`optimal.` → `optimal`, `promoted.` → `promoted`).

One thing your pseudo-code didn't state that I recorded in the comment: for a nested pair this necessarily weakens the enumeration guard, because containment is the only evidence available — `optimal | suboptimal` now reads as `suboptimal` rather than ambiguous. That's the right trade (the guard's live case is a rubric's own *bracketed* template line, which the bracket strip already removes, so the loss is hypothetical while the defect was systematic), but it needed writing down so it isn't "fixed" back. `G21l` now pins both halves — see `72e56c5`.

---

scripts/eval/l1.mjs
mthines: Addressed in `72e56c5` — three checks rather than one, since the nesting has two live shapes and two directions:

- a substring inside a word (`optimal` in `suboptimal`) **and** a hyphenated prefix (`promoted` in `not-promoted`);
- the *contained* choice is still reachable, not shadowed by the longer one — this guards the opposite defect the fix could introduce;
- the documented residue (an unbracketed enumeration of a nested pair resolving to the longer choice), so it stays a known trade.

Proven to bite: reverting the fix reds the first and third. The second stays green by design — the old code didn't have the shadowing defect.

---

scripts/eval/telemetry.mjs
mthines: Addressed in `68515fe`. `AbortSignal.timeout(POST_TIMEOUT_MS)`, 10 s, hoisted to a named constant with the worst case stated (two posts per flush → 20 s). An abort surfaces as an ordinary export failure: caught by `flush()`, stderr only, never an eval failure — which is the module's existing policy, so this makes the timeout case obey it rather than adding a new one.

---

CLAUDE.md
mthines: Addressed in `c7fe3cd`. The row names all four harness files, states the fail-open rationale (one of them can alter any suite's answer, so widening costs tokens while narrowing costs coverage and reports green either way), and names `telemetry.mjs` as the deliberate exclusion — otherwise the next reader re-derives that omission as an oversight and adds it.

Same commit also corrects the heading row above it, which still said `rubric.section` only after this branch merged the `rubric.sections` form.

---

scripts/eval/telemetry.mjs
mthines: Addressed in `6ef109d` — removed. `nowNs` covers every timestamp the exporter emits.

---

scripts/eval/select-suites.mjs
mthines: Addressed in `2ff60da`. Added `detection: boolean` with a one-line note that it is not a `suites` entry and is read as its own job condition — the reason it was the field most worth documenting and the one the type omitted.

---

mthines:
## `bug-detection` is red, and I am standing down on it

The `run-evals` label fired the first full opt-in run ([34088674963](https://github.com/mthines/agent-skills/actions/runs/34088674963)). **Nine suites green, `bug-detection` failed, and `l2` failed with it** — so the required check is red.

**Failing check:** `bug-detection` → `recall@class 60% < 70%`, `fp_rate 40% > 20%`, and a negative verifier lift (−10 pts).

**Why it is not this PR's to fix.** Those rates measure `agents/pr-reviewer/rules/finders.md` and `agents/shared/rules/finding-verifier.md`. Neither is touched here. The only change this PR makes to `scripts/eval/l2-detection.mjs` is the six-line `EVAL_REQUIRE_KEY` guard — the gate terms, the golden set, and the two rubrics are all as they stand on `main`. What this PR changed is that the eval **runs at all**: it previously executed in no workflow, so only its `--self-test` was ever exercised. The red is the eval working on its first permitted measurement, not a regression introduced here.

**No fix exists to port.** Raising the core's precision means editing the two rubrics this eval measures, which per the maintenance table this PR adds needs its own golden decoys and its own re-run — a separate change. Lowering the gate to meet the current core is the fix-to-pass this repo forbids everywhere else, and is precisely what left the eval unrun for as long as it was.

**One thing the run did change, and it is a retraction.** The first detection run showed the verifier halving the finder's false positives (50% → 30%) at *zero* recall cost, and that was written into both `README.md` and `CLAUDE.md` as `finders.md`'s flag-then-filter polarity "measured rather than argued". This run shows it spending 20 points of recall to buy 10 of precision. At 20 seeded records one case is 5 recall points and one control is 10 fp points, so neither estimate survives the other: **the verifier's value is not established.** Both docs now say so, and say that whoever does raise the core's precision must run repeats before *and* after — a single post-change run inside a ±20-point swing cannot tell an improvement from noise. That is a more useful thing to have learned than either number would have been.

No re-run spent to confirm: the push that carries the doc corrections also touches `l2.mjs`, which is a harness file, so selection fans out to all nine suites plus detection on its own.

---

scripts/eval/l2.mjs
```diff
@@ -160,30 +90,89 @@ function rubricFor(suite) {
...
+    "eval.layer": "l2", "eval.model": MODEL,
+    "eval.suite.filter": only, // omitted on a full run — absent means "all suites"
+    "eval.gate.floor": GATE,
+  },
+});
+let totalCases = 0, totalPass = 0, totalInTok = 0, totalOutTok = 0, totalCacheRead = 0, totalCacheWrite = 0;
+
 for (const suite of SUITES) {
   if (only && suite.name !== only) continue;
   const goldenPath = join(REPO_ROOT, "scripts/eval", suite.golden);
```
mthines, line 248: issue (high): 🟠 **A selected suite with no golden cases exits 0** **(non-blocking)**

Deleting or emptying a golden file makes its suite print `(skip …)` or `0/0 (NaN%) [advisory]` and exit 0 — the third door of the class this PR closes for unknown suites and absent keys.

Evidence: `scripts/eval/l2.mjs:179` (deleted golden skips) · `scripts/eval/l2.mjs:254` (empty golden reports advisory)

```js
if (!cases.length) {
  console.error(`✗ L2: ${suite.name} was selected but has no cases.`);
  process.exit(4);
}
```

<sup>`pr-reviewer` · commit `621489f`</sup>



---

.github/workflows/evals-l2.yml
```diff
@@ -54,13 +225,139 @@ jobs:
...
+          echo "::warning title=Detection eval skipped::No ANTHROPIC_API_KEY secret — the eval cannot run. Add the secret under Settings → Secrets and variables → Actions."
+      - name: L2 bug-detection eval
+        env:
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          # Gated, unlike the report-only default: the two rates ARE the contract this
+          # eval exists to hold, and a detection core that stops finding bugs while
+          # reporting green is the failure it was written for.
+          EVAL_DETECTION_GATE: "1"
+          EVAL_REQUIRE_KEY: "1"
+          OTEL_EXPORTER_OTLP_ENDPOINT: ${{ vars.OTEL_EXPORTER_OTLP_ENDPOINT }}
```
mthines: suggestion (medium): 🟡 **Detection job sets OTLP env it cannot use** **(non-blocking)**

`l2-detection.mjs` imports only `node:fs`, `node:path` and `lib.mjs` — no exporter, no spans — so these three variables, one of them the Dash0 token, are inert.

Evidence: `scripts/eval/l2-detection.mjs:36` (all three imports)

<sup>`pr-reviewer` · commit `621489f`</sup>



---

agents/pr-reviewer/rules/depth-routing.md
```diff
@@ -67,7 +67,7 @@ otherwise claim it and buy a full lens pass to re-read the reviewer's own asks.
 
 **One exclusion runs before the table.** A delta whose shapes are exclusively `docs-only`,
 `test-only`, or generated, **and** whose `blast_radius.band == none`, does not consider the size
-triggers at all — neither `DELTA_LINES > 100` / `NEW_FILES > 0` in the `deep` row nor the
+triggers at all — neither `D12` / `D13` in the `deep` row nor the
```
mthines: suggestion (low): ⚪ **Exclusion cites D12/D13 before they are defined** **(non-blocking)**

The checklist begins at line 89 and defines these two at 100–101, so a reader meeting the exclusion first has no referent for them. The prose this replaced named the conditions outright.

Evidence: `agents/pr-reviewer/rules/depth-routing.md:100` (D12 defined here)

```markdown
triggers at all — neither the `DELTA_LINES > 100` nor the `NEW_FILES > 0`
trigger (`D12` / `D13` in the `deep` row below) nor the
```

<sup>`pr-reviewer` · commit `621489f`</sup>



---

CLAUDE.md
```diff
@@ -322,11 +322,39 @@ All applicable hops must resolve. If any is missing, the harness will not see th
...
...case.duration` (Histogram, `s`) and the upstream-semconv `gen_ai.client.token.usage` (Histogram, `{token}`, split by `gen_ai.token.type`). Four rules, all guarded by `G21k` and each probed by breaking it: **off unless `OTEL_EXPORTER_OTLP_ENDPOINT` is set** (a fork or fresh clone attempts no export); **a MISS is not a span error** — a wrong answer is the measurement, so the case span stays `UNSET` and only a transport/API failure sets `ERROR`, since conflating them makes every rubric regression look like an outage in the trace list; **an absent attribute is omitted, never a placeholder** (`unknown` is not queryable as absent); and **the flush precedes the gate exit**, so a run that trips `EVAL_GATE` still ships the trace you most want to read — asserted positionally, because a refactor that moves the exit up would lose exactly that trace. Export failure never changes the verdict: the accuracy is the product, the span is the receipt. In CI the endpoint is a repo **variable** (a hostname is not a secret) and only `DASH0_AUTH_TOKEN` is a secret; each matrix job is its own process, so an opted-in PR yields **one trace per suite** — group by the `cicd.pipeline.run.id` resource attribute.
```
mthines: suggestion (low): ⚪ **Scorer-rule count stale in the commit that staled it** **(non-blocking)**

This bullet says three rules and lists three, while the same commit adds a fourth to `G21l` — nested-choice containment plus its by-design residue, which the runner's own comment says not to undo.

Evidence: `scripts/eval/l1.mjs:277` (the fourth rule)

<sup>`pr-reviewer` · commit `621489f`</sup>



---

scripts/eval/l2.mjs
```diff
@@ -193,35 +182,153 @@ for (const suite of SUITES) {
...
+// distinction matters: three of nine suites reported MISSED purely because their rubric
+// slice is small, which sent a reader looking for a bug that does not exist. Naming the
+// likely cause costs one line and saves that hunt.
+const MIN_CACHEABLE_TOKENS = 1024;
+
+if (totalInTok || totalOutTok) {
+  let cacheNote;
+  if (totalCacheRead || totalCacheWrite) {
+    cacheNote = ` · cache ${totalCacheRead.toLocaleString()} read / ${totalCacheWrite.toLocaleString()} written`;
+  } else if (totalCases && totalInTok / totalCases < MIN_CACHEABLE_TOKENS) {
```
mthines: suggestion (low): ⚪ **Cache heuristic measures the prompt, not the prefix** **(non-blocking)**

`cache_control` marks the system block alone, but `input_tokens` also covers each case's user message — so a sub-1024-token rubric with large cases still reports the alarming `cache MISSED`.

Evidence: `scripts/eval/l2.mjs:112` (only system is cached)

<sup>`pr-reviewer` · commit `621489f`</sup>



---

mthines (commented):
Reviewed `621489f` — 5 findings — 1 high, 1 medium, 3 low; nothing blocking, and every gate passes. 5 finding(s) inline. A sticky report already exists on this PR, but this access path (GitHub MCP, no `gh`) exposes no tool that can edit an issue comment — so the report was not updated in place and is in the run's terminal output instead.

---

scripts/eval/l2.mjs
mthines: Addressed in `f3d78e4` — two guards, because the two shapes fail differently and only one of them was the one you could see. A missing golden file printed `(skip …)` and moved on; an *emptied* one ran zero cases, scored `NaN%`, and could not breach the floor either, because `results.length >= GATE_MIN_CASES` is false at zero — so the suite appeared in the summary looking graded. Both now exit 4, with a message naming the remedy (delete the `SUITES` entry too, or restore the file).

Proven to bite by execution rather than by reading, with a dummy key so the guard is reached ahead of any API call: pre-fix, deleting `golden/tier-routing.jsonl` gave `(skip tier-routing: no golden file)` and exit 0, and emptying it gave a printed summary and exit 0. Post-fix both exit 4. Golden file restored and verified byte-clean afterwards.

The framing is yours and it is the right one — this was the third door of the same class, and it was the worst of the three, since the other two at least stay quiet about having measured nothing.

---

.github/workflows/evals-l2.yml
mthines: Addressed in `ae2c56f` — all three removed, verified by parsing the workflow rather than by eye: the detection job's env is now exactly `ANTHROPIC_API_KEY`, `EVAL_DETECTION_GATE`, `EVAL_REQUIRE_KEY`, and the `suite` job still carries all three OTLP variables.

The token half is the part worth naming, and I left a comment saying so: `DASH0_AUTH_TOKEN` was being placed into an environment nothing reads, which is a credential in a process with no use for it rather than merely dead config.

The comment also records what the removal is *not*: trending this eval's two rates is worth doing — the first two runs disagree by 15 recall points, which is precisely the question a backend answers and stdout cannot — but that needs a telemetry module in the runner first. The env goes back when it has one, rather than sitting there implying it already does.

---

agents/pr-reviewer/rules/depth-routing.md
mthines: Addressed in `b10388a`, taking your approach — gloss the two inline where the exclusion cites them (`**D12**` `DELTA_LINES > 100` and `**D13**` `NEW_FILES > 0`) rather than moving the paragraph below the checklist.

Two reasons for inlining over reordering, both about this specific file. It is `shape-depth-routing`'s live rubric, so a moved block changes more of the prompt than two added definitions do, and the suite is at 20/22 — not the moment to perturb the sample beyond the fix. And `G38h` asserts both pre-table rules sit above the table; keeping them adjacent and in their documented order leaves that reading intact instead of relying on the checklist landing between them.

I also tried adding a waiver pointer under the checklist's lead-in and reverted it: line 108 already carries that exact sentence, two lines below the table it belongs to.

This is the second finding of yours on this file that was about *reachability of a stated rule* rather than the rule itself — same shape as the thirteen-condition cell, one layer up.

---

CLAUDE.md
mthines: Addressed in `8b574af` — now four, with the fourth written out rather than just counted: nested-choice containment resolving to the longest match, the two-direction probe that keeps the contained choice reachable, and the by-design residue (an unbracketed enumeration of a nested pair resolving to the *longer* choice instead of reading as ambiguous), noted as a deliberate trade that cannot be "fixed" back by re-breaking the line above it.

Worth stating why this one mattered more than a stale number: this bullet is the audit surface for `G21l`. A reader counting the live checks against it would have found one more guard than the prose accounts for, with no way to tell an undocumented guard from a stale count — and the check they'd have been unable to place is the one whose whole purpose is to stop a future author restoring the defect.

---

scripts/eval/l2.mjs
mthines: Addressed in `4dee60c`. The note now measures the **prefix**: the run tracks `maxSystemChars`, the largest system block it built, and asks whether *that* clears the bound — so a small rubric with large cases can no longer be dragged over it by user-message tokens and cry `MISSED`, which is exactly the inversion you describe.

Your diagnosis was better than the fix I'd already shipped. The earlier commit separated `cache not applicable` from `cache MISSED` but decided between them on total input ÷ case count, which is the wrong quantity for the reason you give — only the system block carries `cache_control`. The message was right and the arithmetic behind it was still measuring the whole prompt.

One deliberate asymmetry, recorded in the comment: chars/token is an estimate and runs ~10% low against the observed writes, so a prefix within ~15% above the bound reads as "not applicable". That errs toward a false *"nothing to fix"* on a borderline suite, which costs a missed discount, rather than a false `MISSED`, which costs a reader a hunt for a defect that does not exist — the failure being fixed here.

Same commit corrects a claim in `ask()`'s header that this measurement refuted: three suites fall under the bound, not two, and they are not all cheap. `tier-routing` is ~935 tokens — it misses by under 90 while being the largest uncached bill in the run (30 cases × ~1k ≈ 30k input tokens, more than any cached suite pays *after* its discount). So "nothing is lost when they miss" was false for that one. Padding a rubric to reach the bound would be writing for the biller instead of the reader, so the loss is named and accepted rather than engineered away.

---

mthines:
**Standing down on the one red check, deliberately.** Recording it here rather than leaving it to be inferred from the description.

**Failing check:** `evals · L2 (behavioral) / bug-detection`, which fails the `l2` aggregator with it. Nine suites are green; `l2` is red *because* it correctly reads `needs.detection.result`, which is the behaviour `G21m` exists to guard.

**Why it is not this PR's.** `git diff origin/main...HEAD -- scripts/eval/l2-detection.mjs` is a 6-line `EVAL_REQUIRE_KEY` guard. `agents/pr-reviewer/rules/finders.md`, `agents/shared/rules/finding-verifier.md` and `golden/bug-detection.jsonl` — the three inputs the eval measures — are untouched on this branch, and the `recall ≥ 0.7` / `fp ≤ 0.2` gate is pre-existing on `main`. The eval simply ran here for the first time: before this PR it was wired into no workflow, so only its `--self-test` had ever executed.

**No fix exists to port.** Getting it green means editing the two rubrics it measures, which per this PR's own maintenance table owes new golden decoys and its own repeated re-runs — a change with its own diff, not a rider on this one. Lowering the gate to meet the current core is the fix-to-pass this repo forbids everywhere else, and is precisely what left the eval unrun to begin with.

**Not a flake, and no re-run spent to establish that.** It has now failed four consecutive times with a consistent shape: `fp_rate` after the verifier at 30 / 40 / 30 / — never once at or under 20 — while recall straddles its 70% bar. A rate that reproduces across four independent samples is not transport noise, so a fifth run would buy nothing.

Full numbers, the run-2 retraction, and where the work goes (the controls, not recall) are in the description and in `scripts/eval/README.md`.

---

scripts/eval/golden/code-review-retrieval-relevance.jsonl
```diff
@@ -3,3 +3,12 @@
...
+{"id":"codebase-knowledge-tag-outside-read","input":"PR diff (repo mthines/agent-skills): changes scripts/eval/lib.mjs extractSection. Candidate memory: {tag: 'codebase-knowledge', scope: 'repo::mthines/agent-skills', source.agent: 'pr-reviewer', key: 'knowledge::extractSection@scripts/eval/lib.mjs', gist: 'extractSection is heading-level-aware, so a parent heading captures every deeper subsection under it'}","expected":"skip","notes":"BOOTSTRAP SEED — NOT A REAL BASELINE. Ground truth = the tag filter. codebase-knowledge is one of this agent's own buckets and this record was written by pr-reviewer, but it is read at Step 1.2a (impact-keyed) rather than by either path in scope here: Step 1.0's four calls pass tags=['loop::reviewer-lessons'] or tags=['loop::reviewer-comment-relevance'], and Step 1.2c keeps ONLY returned hits carrying one of those two tags. Harder than fix-bug-lesson-wrong-loop-tag, where the tag belongs to a different loop entirely — here every other field is in range and the record is exactly on-topic for the diff."}
```
mthines, line 14: issue (high): 🟠 **Retrieval suite crossed into gating and is red** **(blocking)**

Growing the set from 5 to 14 crosses `EVAL_GATE_MIN_CASES`, so it gates instead of reporting `[advisory]`. CI at this head scores it 9/14 (64%), under the 70% floor, and exits 1.

Evidence: `scripts/eval/golden/code-review-retrieval-relevance.jsonl:14` (14 cases) · `scripts/eval/README.md:537` (still documented advisory)

<sup>`pr-reviewer` · commit `e65c302`</sup>



---

scripts/eval/l2.mjs
```diff
@@ -160,68 +117,381 @@ function rubricFor(suite) {
...
+// slice is small, which sent a reader looking for a bug that does not exist.
+//
+// Measure the PREFIX, not the prompt. Only the system block carries `cache_control`
+// (see `ask()`), while `input_tokens` also covers each case's user message — so
+// dividing total input by case count overstates the cached part and can cry MISSED at
+// a suite that was never cacheable. `maxSystemChars` is the largest system block the
+// run built: if even that is under the bound, no suite in the run could cache.
+//
+// Chars/token is an estimate and runs ~10% low against the observed writes, so a
+// prefix within ~15% above the bound reads as "not applicable". That is the direction
```
mthines: issue (medium): 🟡 **Documented 15% cache tolerance is not implemented** **(non-blocking)**

The comment promises that a prefix within ~15% above the bound reads as not applicable, but the test is a strict `prefixTok < 1024`, so 1024-1177 still print `cache MISSED`.

Evidence: `scripts/eval/l2.mjs:452` (the promise) · `scripts/eval/l2.mjs:473` (the test)

```js
} else if (maxSystemChars && prefixTok < MIN_CACHEABLE_TOKENS * 1.15) {
```

<sup>`pr-reviewer` · commit `e65c302`</sup>



---

scripts/eval/README.md
```diff
@@ -158,6 +458,273 @@ floor (or switch to 0%-regression-vs-baseline) and gate it hard.
...
+shows a `deep` and a `quick` and **no `standard`** — the tier the golden set calls "the
+middle tier's default population".
+
+**3. The variance is the third finding.** Same rubric, two runs: `shape-depth-routing`
+20/22 → 18/22, and `tier-routing` held at 5 misses while the *set* changed
+(`lite-error-toast` in, `full-api-refactor` out). At n=22, ±2 cases is ±9 points. The 70%
+catastrophic floor survives that; a "no regression vs baseline" gate would be pure noise
+at these sizes. Do not tighten a floor until a golden set reaches ≥ 50.
+
+**4. `code-review-retrieval-relevance` should not have been gating.** 5 cases at a 70%
```
mthines, line 543: issue (medium): 🟡 **README still calls the retrieval suite advisory** **(non-blocking)**

Finding 4 asserts the suite `is now [advisory]` and the run table repeats it, but this merge takes it to 14 cases, which `CLAUDE.md` correctly says gates.

Evidence: `scripts/eval/README.md:537` (is now advisory) · `scripts/eval/README.md:599` (run table) · `CLAUDE.md:325` (says it gates)

<sup>`pr-reviewer` · commit `e65c302`</sup>



---

scripts/eval/README.md
```diff
@@ -158,6 +458,273 @@ floor (or switch to 0%-regression-vs-baseline) and gate it hard.
...
 checks" → pick **evals · L1** (and **evals · L2** once the secret is set).
 
+### The L2 baseline — first real run
+
+The suites existed for 232 CI runs without the `ANTHROPIC_API_KEY` secret, so every
+one of those greens measured nothing. This is the first run that actually executed,
+recorded here so the next change has something to compare against rather than a
+remembered number.
+
+`claude-sonnet-4-6` · `EVAL_GATE=70` · nine suites, 140 cases · **130/140 = 92.9%**:
```
mthines, line 468: suggestion (low): ⚪ **Nine-suite baseline counts stale after the merge** **(non-blocking)**

The baseline still reads `nine suites, 140 cases` with the retrieval row at 5, and `CLAUDE.md` repeats `130/140` and `131/140`; the live nine-suite total is 149.

Evidence: `scripts/eval/README.md:468` (140 cases) · `scripts/eval/README.md:480` (retrieval row) · `CLAUDE.md:325` (130/140 and 131/140)

<sup>`pr-reviewer` · commit `e65c302`</sup>



---

mthines (commented):
Reviewed `e65c302` — 4 findings — 1 blocking: the retrieval suite crossed into gating and is red at 64.3%. 4 finding(s) inline. The GitHub MCP access path exposes no issue-comment update tool, so the existing report comment could not be rewritten in place; the full report is in the run's terminal output.

---

scripts/eval/l2.mjs
mthines: Addressed in `235f6ed`, but by fixing the **comment** rather than the code — your patch is the one change I deliberately did not make, and the reason is that the tolerance was reasoned in the wrong direction.

Your read of the contradiction is exactly right, and your execution numbers reproduce (1023 → `not applicable`; 1024/1030/1100/1177 → `cache MISSED`). But applying `* 1.15` would have made the report worse. The comment's own premise is that chars/token "runs ~10% low against the observed writes" — so an estimate that *already* clears 1024 implies a real prefix comfortably above it. That suite genuinely could have cached, and `MISSED` is the true report. Widening the band upward would have suppressed true `MISSED` reports, which is the opposite of what the note exists to protect.

The uncertainty is entirely on the other side: an estimated ~950 may be a real ~1050, so `not applicable` is already the lenient verdict and is the one that can be wrong. The comment now says that, and names the residue one way — a prefix measured just *below* the bound may have been cacheable and we will have said "nothing to discount" (costs a missed discount), while a `MISSED` report is trustworthy and should be acted on.

Behaviour is unchanged, which matters here: `l2.mjs` is a harness file, so a behavioural edit would owe all nine suites plus detection a re-run to prove no label moved. A comment-only fix owes nothing.

Worth noting where this came from: the 15% claim entered in `4dee60c` as part of fixing your *earlier* prefix-vs-prompt finding on this same block. That fix was right about the quantity to measure and wrong about the tolerance, and it shipped the correct code with an incorrect justification attached — which is the harder half to catch, since the code was never broken.

---

scripts/eval/README.md
mthines: Addressed in `213bc19`. The present-tense claim is the defect and it is gone: a superseding note now records that the decoy set took the suite back over `EVAL_GATE_MIN_CASES`, that it therefore grades for real, and that it measured **9/14 (64.3%)** at `e65c302` — below the floor.

One deliberate departure from the literal ask. I did **not** rewrite the historical run figures (`3/5 … [advisory]` in the run table, and the row in the first-baseline table). Those are an accurate record of runs that really did execute a 5-case set, and restating them as 14 cases would falsify the record to make the page self-consistent. What was actually missing is the warning that the population changed underneath them, so the note carries it and says explicitly that no figure for this suite predating the decoys is comparable to a post-decoy run, with a pointer to `NOTES.md` for the per-case history.

You were right that this was the sharper of the two doc findings — `CLAUDE.md` and `README.md` disagreeing on whether a gate gates is worse than either being stale, because a reader resolving the conflict has no way to tell which one the CI actually obeys.

---

scripts/eval/README.md
mthines: Addressed in `9e59b36`. Your 149 reproduces (15 + 18 + 14 + 14 + 15 + 6 + 15 + 22 + 30).

Same treatment as the sibling finding: both aggregates stay, each being correct for the run that produced it, and what they gain is a stamp saying the denominator moved. `CLAUDE.md`'s L2 bullet now says so inline next to `131/140`, and the README baseline header carries the longer form — nine suites are 149 cases now, so compare per suite and never on the aggregate.

Leaving the numbers and marking them is the right shape for this page specifically, because the passage two sentences earlier already warns "never read the total as a quality signal" and then hands the reader a total to read. The stamp closes that gap rather than papering over it: the composition was always the thing to read, and a changed case count is one more reason the aggregate cannot be compared across time.

---

scripts/eval/golden/code-review-retrieval-relevance.jsonl
mthines: **Confirmed, not fixed, and deliberately left open — this one needs a human call.** Your framing is right and the `G21n` observation is the sharpest part of it: that guard asserts the *majority-class baseline* sits below the floor, not that measured accuracy clears it, which is exactly why L1 stays green at 1612 while the suite gates red.

**Where the failure lives.** It is not the merge resolution. It reproduces on #183's own head — `l2` there is red on the same suite at **8/14 (57.1%)** with an overlapping miss set — so the decoy set arrived already red and crossing `EVAL_GATE_MIN_CASES` made it *visible* rather than making it fail. This head's 9/14 is one case better than #183's, which is within the ±1–2 variance this page documents.

**Diagnosis — the judge prompt asks a different question than the labels encode.**

| | says |
|---|---|
| golden `notes` | ground truth is **list-reachability** — "the lesson still surfaces via list-reachability" |
| suite `instruction` | does the read "surface the candidate record … **to the finders**" |
| rubric §1.0:1024 | "such a record does not **reach the finders**" |
| rubric §1.0:1026 | a diff-unrelated lesson is in the pool this step loads and "stops being a candidate only downstream" (1.2d) |

Under "to the finders", `skip` is the **correct** answer for a diff-unrelated lesson — the rubric says so in those words, and 1.2d is not even in the extracted rubric. So the instruction contradicts the labels through the rubric's own vocabulary. All five misses are `surface→skip` while 6/6 skip decoys pass, which is the signature of a downstream relevance filter being applied rather than a broken rubric.

This is also a repeat of a defect class already fixed here once: the instruction's verb was changed (`would be surfaced … for the given PR diff` → `surfaces`) to remove precisely this relevance question, and `to the finders` was left in place, reintroducing it one clause over.

**Why I did not apply it.** Editing a judge prompt is gated at `confidence(analysis) ≥ 90%` in this repo. I scored it and got **80%** (evidence 88 / root-cause 78 / outcome 72). Two things hold it down, and neither is rhetorical:

1. It explains 4 of 5 misses. `seen2-below-threshold` is diff-*related*, so this mechanism does not account for it, and its scoped twin flipped between #183's run and this one.
2. A live competing hypothesis: §1.0 **already** says "narrowing this read by your own sense of what looks relevant to the diff is a defect, not an optimisation" — and the model narrowed anyway. If emphasis did not move it, re-framing the question may not either.

The measurement that would settle it is running the suite with the instruction changed, and that needs an `ANTHROPIC_API_KEY` the sandbox does not have. Applying the edit and letting CI referee would be using the gate as a coin-flip on the one file class this repo says never to fix-to-pass.

**Proposed patch, for whoever takes it** (belongs in #183 with `suites.mjs`, or here if you want #182 green first):

```diff
-decide whether that read surfaces the candidate record described to the finders.
+decide whether that read RETURNS the candidate record into the pool Step 1 loads.
+Whether it survives the later diff-keyed shortlist is a different question and not this one.
```

It also makes the question answerable *from the supplied rubric* — "reaches the finders" depends on §1.2d, which the suite does not extract, and this repo already treats a question broader than its rubric as a defect.

The three sibling findings are fixed and resolved (`235f6ed`, `213bc19`, `9e59b36`). This thread stays open on purpose: the PR is red, and no reading of it should suggest otherwise.

---

mthines:
## Status at `9e59b36` — one red suite, and it is not this PR's defect

Correcting two claims in the PR description that the #183 merge invalidated, and recording why a red check is being left red.

**The description is stale in two places.** It says "New L1 checks (**1587 total, green**)" — L1 is now **1612/1612 green**. And its "Eval obligation — discharged, twice" section reports "9/9 suites green", which described the pre-merge head. At `9e59b36` the picture is:

| check | state |
|---|---|
| `l1` | ✅ 1612/1612 |
| 8 of 9 L2 suites | ✅ green |
| `suite (code-review-retrieval-relevance)` | ❌ **9/14 (64.3%)** — below the 70% floor |
| `bug-detection` | ❌ red on the pre-existing detection core, as the description explains |

The body itself is left unedited rather than rewritten in place: the description refresh belongs at convergence, and this PR has not converged (one blocking thread open by design). Treat this comment as authoritative over the body's result figures until then.

**Why the retrieval suite is red, and why I did not make it green.** #183's decoy set grew it 5 → 14 cases, taking it back over `EVAL_GATE_MIN_CASES` (10), so it grades instead of reporting `[advisory]`. It arrived already failing — #183's own `l2` is red on the same suite at **8/14 (57.1%)** with an overlapping miss set. Crossing the threshold made the failure *visible*; it did not cause it. Worth noting `G21n` asserts the majority-class *baseline* sits below the floor, not that measured accuracy clears it, which is why L1 stays green while the gate fails.

The cause looks like the suite's judge prompt asking a different question than its labels encode — `instruction` says "surfaces … **to the finders**", the golden `notes` say list-reachability, and rubric §1.0 uses "reach the finders" for the *downstream* drop. Under that wording `skip` is correct for the diff-unrelated cases, which is the signature: all five misses are `surface→skip` while 6/6 skip decoys pass.

**The fix is proposed, not applied,** because editing a judge prompt is gated at `confidence(analysis) ≥ 90%` here and it scored **80%** — the mechanism explains 4 of 5 misses, and §1.0 *already* forbids diff-narrowing in as many words yet the model narrowed anyway, so re-framing may not move it either. Settling that needs a run with the instruction changed, which needs an API key this environment does not have. Applying it and letting CI referee would be fix-to-pass on the one file class this repo says never to treat that way. Patch and full reasoning: [the open thread](https://github.com/mthines/agent-skills/pull/182#discussion_r3949022662).

**Fixed this round** — the review's other three findings, one commit each:

- `235f6ed` — the cache note promised a ~15% tolerance the code never implemented. Fixed the **comment**, not the code: the tolerance was reasoned in the wrong direction (a low-running chars/token estimate means an estimate clearing 1024 implies a real prefix above it, so `MISSED` is true there and the uncertainty sits *below* the bound). Behaviour unchanged, so no label can move.
- `213bc19` — `README.md` called the retrieval suite `[advisory]` while `CLAUDE.md` correctly said it gates. Historical run figures kept as an accurate record of 5-case runs; a superseding note carries the current state and says none of them is comparable.
- `9e59b36` — `130/140` and `131/140` stamped with the moved denominator (nine suites are **149** cases now).

---

PR #183, opened 2026-09-06
Make the retrieval-relevance suite diagnostic, and its rubric self-consistent

mthines:
Follow-up to #181, which merged with `l2` deliberately red. This is the deferred half.

## The problem

`code-review-retrieval-relevance` sat at 3/5 (60%) against a 70% floor, red identically on `main`. Two hypotheses were already tried and **both falsified**:

| # | Hypothesis | Fix shipped | Result |
| --- | --- | --- | --- |
| 1 | The suite was fed a rubric 2.5× broader than its own question (`## Step 1`, 67,630 chars, vs. the two subsections that own it, 27,568) | `rubric.sections` re-pointed to `### 1.0` + `### 1.2c` | run 34044758211 — **3/5, same two misses** |
| 2 | Step 1.0 never stated its read's *non*-filters | Added them, including *"Narrowing this read by your own sense of what looks relevant to the diff is a defect, not an optimisation"* | run 34044924487 — **3/5, identical** |

Both fixes were real defects and both stand. Neither moved the eval. **The model read hypothesis 2's sentence and still answered `skip`** — which is what localises the remaining problem to the *question*, not the rubric.

Two things were wrong with the question, and they had to be fixed together:

- **The suite could not grade what it claimed to.** Labels ran 4 `surface` / 1 `skip`, so an always-`surface` responder scored **80%** and cleared the floor without reading the rubric. A green meant only "the model stopped answering `skip`".
- **The `instruction` asked a relevance question.** *"…would be surfaced by the documented read **for the given PR diff**"* puts the diff in the framing, which reads as an invitation to judge whether the record is relevant to the change — the exact judgement the rubric calls a defect. Both persistent misses turn on that distinction.

Shipping either half alone makes things worse. Disambiguating the question against a 4/1 set hands over the answer key (extremal Goodhart, which the anti-gaming guard forbids); adding decoys under the misleading framing most likely converts two known misses into several.

## What changed

**1. Six `skip` decoys, one per filter dimension of the read under test.** Derived from the rubric rather than invented, so none is answerable without reading it:

| Dimension | Decoy | Why it is hard |
| --- | --- | --- |
| tag | `codebase-knowledge` | This agent's own bucket, written by `pr-reviewer` — but read at Step 1.2a, by neither path in scope. Every other field is in range and the gist is exactly on-topic. |
| scope | `branch::…::feat/suppression-order` | Right tag, high `seen_count`, gist directly relevant. |
| scope | `repo::mthines/lorekit` | Right prefix, wrong repo, with a deliberately on-topic gist — relevance cannot admit an out-of-scope record. |
| expiry | `expires_at: 2026-07-01` | Right tag, right scope, the highest `seen_count` in the set, gist matches the diff exactly. |
| source | `source.agent: aw-executor` | Another tool's record about the same repository, `source.explicit: false`. |

Plus `global` and `source.explicit: true` as **positive** counterparts, so scope and attribution are tested in both directions rather than only as rejections.

**14 cases, 8 `surface` / 6 `skip` — majority-class baseline 57.1%.** Neither degenerate strategy passes the unchanged 70% floor, and the floor now tolerates 4 misses instead of 1.

**2. The two persistently-missed originals were not edited.** They have explicitly-scoped superseding twins stating the `scope` and `source.agent` fields the originals leave silent; the originals stay runnable as regression guards, per `eval-iterate`'s no-overwrite-in-place rule.

**3. The `instruction` asks a procedure-application question, and its verb is `surfaces`** — matching the labels the suite has always used. This is not cosmetic: Step 1.0's source-attribution filter runs on what its four calls *returned*, so a record another tool wrote **is** returned and then dropped. Under "returns" the two attribution cases were answerable both ways and graded nothing.

**4. Fixing (3) exposed a self-contradiction in the live rubric.** `### 1.0` said its four calls *"filter on three things and nothing else — tag, scope, and expiry"* (added by hypothesis 2, in this same series) while **rule 1 of the same section** binds `source.agent == "pr-reviewer" ∨ source.explicit == true` on *"every read this step and Step 1.2c issue"*. Both decoys above were graded on a rubric that contradicted itself on their own dimension. Resolved by naming attribution as the fourth filter and stating *where* it runs. The same edit states that the `scope` parameter matches **exactly** — verified as `m.scope = p_scope` in lorekit migration `00108`, not asserted from memory — which is why `memory.md` can park per-PR state in a `branch::` scope and rely on it staying out of the lesson read.

**5. New L1 guard `G21h`.** Asserts the majority-class baseline sits below the `EVAL_GATE` floor, deriving the labels from the JSONL and the floor from `evals-l2.yml` rather than re-encoding either. Re-degenerating the split now reds L1 instead of only a paid L2 run. It **fails closed** when no `EVAL_GATE` literal is found: an unknown floor cannot be shown to exceed a baseline.

Docs updated in the same commit: `NOTES.md` (its `skip` ground truth named only the tag dimension, so the other three had no definition and the set tested none of them), `scripts/eval/README.md`, and `CLAUDE.md`'s two recorded limitations — rewritten to describe what holds now, not edited to pretend they never applied.

## Verification

- **L1: 1519/1519** (was 1518; `+1` is `G21h`).
- **`G21h` bite-tested, both branches, expectations declared before running.** Flipping two `skip` labels to `surface` (10/4 = 71.4%) → one red, `G21h`, reporting `majority-class baseline 71.4% >= gate 70% … (n=14, skip=4 surface=10)`. Deleting the `EVAL_GATE` line → one red, `G21h`, via the fail-closed message. Flipping only one label gives 64.3% and correctly stays green, so two is the minimum that bites.
- **Every claim in all nine new records checked against the live rubric text**, extracted through `extractSection` rather than recalled: the four `memory_list` calls and their two scope values, `order="rank"` on the global pair, `view="summary"` returning the index and not bodies, `"Skip expired entries"` on both paths, and the `source.explicit` disjunct verbatim.
- `l2` **cannot run locally** — no `ANTHROPIC_API_KEY` in this environment (a direct probe returns `401`). Every confirmation for this suite goes through CI.

## Stated limitations

- **The anti-gaming gate did not clear.** `confidence(analysis) = 86%`, below the 90% bar, failing on root-cause certainty: this is the third hypothesis and two are already falsified. The `critical` pass that gates it found the two must-fixes in (3) and (4), which are fixed here. **The eval-definition half of this PR is provisional pending CI evidence, not a confirmed fix.** The rubric fixes in (4) stand on their own — they are fixes to the thing under test, not to the eval. If CI returns 5-of-5 (the confirmation bar for a judge-graded path), root-cause certainty stops being a prediction and the gate clears on evidence; if it returns red, hypothesis 3 is falsified like the other two, in writing rather than behind a green badge.
- **The margin is thin, and thin margins punish a judge-graded path.** The floor tolerates 4 misses of 14. If both persistent `surface` misses survive *and* two decoys miss, the suite passes at 71.4% and fails at 5 — and 5 consecutive passes is a much harder bar than one at that margin.
- **`out-of-scope-branch-scope` needs a fact the rubric states only as of this PR.** The label is factually right (verified in SQL), but before change (4) the model had to infer exactness from the fenced block's two literal scope values.
- **`codebase-knowledge-tag-outside-read` is decidable only via the instruction's `ONLY … below` scoping** — the record *is* surfaced to the reviewer, at Step 1.2a. Deliberately the hardest tag case; it grades scope-reading discipline as much as tag-filtering.
- **Nine of fourteen cases are now authored by the same agent that read the rubric closely.** Author bias toward "the rubric says X, so the answer is X" is unmeasurable from inside this run. The `BOOTSTRAP SEED — NOT A REAL BASELINE` marker on every record is a label, not a control.

## Not done

Nothing was loosened. No original case was edited, skipped or deleted; `EVAL_GATE` is untouched at 70; no L1 guard was weakened; no CI step was disabled or made `continue-on-error`. Adding passing cases to cross the floor was considered and refused — it is the textbook extremal Goodhart the guard exists to stop.

The other eight L2 suites are unaffected by the `rubricFor` change (run 34044749700, nine-suite pass): lowest is `shape-depth-routing` at 81.8%, all clear the floor.

https://claude.ai/code/session_0134kyioxSrXMMz4KwZDm7QR

---
_Generated by [Claude Code](https://claude.ai/code/session_0134kyioxSrXMMz4KwZDm7QR)_


mthines:
## `l2` red on `f240418` — every request was rejected, so the run measured nothing

[Run 34048141782](https://github.com/mthines/agent-skills/actions/runs/34048141782/job/101526793120) scored **0.0% on all nine suites** — 189 cases, every one `ERR(API 400: {"type":"error","erro`. It exited with *"a suite is below the `EVAL_GATE` floor of 70%"*, which reads as a rubric or golden-set regression. It is not one.

| suite | this run | this branch touches it? |
| --- | --- | --- |
| `tier-routing` | 0/30 | no |
| `bug-class` | 0/18 | no |
| `complexity-triage` | 0/14 | no |
| `aw-should-trigger` | 0/15 | no |
| `optimize-approach-optimality` | 0/15 | no |
| `reviewer-agreement-bump` | 0/6 | no |
| `severity-tiering` | 0/15 | no |
| `shape-depth-routing` | 0/22 | no |
| `code-review-retrieval-relevance` | 0/14 | yes |

**Why this is not this PR's failure.** Six of those suites are untouched by this branch, and the only non-comment change to `scripts/eval/l2.mjs` in the whole diff is one suite's `instruction` string — `tier-routing`'s prompt is byte-identical to `main`'s. A request-shape defect would also have crashed the script rather than producing 189 per-case HTTP 400s. For comparison, [run 34044749700](https://github.com/mthines/agent-skills/actions/runs/34044749700) exercised the same nine suites earlier today and returned real scores (81.8%–100%). A single-suite `workflow_dispatch` on `main` is running now to confirm base-branch redness directly.

**What I have pushed** (`d7d157a`) is not a fix for the 400 — it is a fix for the fact that the 400 was **unreadable by construction**, which is what cost this diagnosis a cycle. `ask()` captured 120 chars of the API's response body, and the per-case report line then sliced the message to 30. The first ~30 chars of an Anthropic error are the JSON envelope — `{"type":"error","error":{"type":…` — so the truncation cut it off immediately before the part that names the cause. The API's explanation entered the process in exactly one place and was discarded there.

- `ask()` now keeps **600** chars; a model-not-found or context-length message can exceed 120 on its own.
- Distinct messages accumulate in a `Map` with a count and are **reprinted in full** at the end, sized by the share of the run they consumed. A blanket transport failure is one entry covering everything; a real per-case problem is a small count.
- A run where **>50% of cases never reached the model** now exits `INVALID RUN`, not "below the `EVAL_GATE` floor", and says in as many words not to read it as a rubric or golden-set regression. Half a run failing to transport is not a measurement.

Self-tested end to end against the real API with a deliberately invalid key, expectations declared before running: 14/14 errored, one distinct message printed in full (`authentication_error`), `INVALID RUN`, exit 1, no `EVAL_GATE` line. L1 stays 1519/1519.

**The 400 itself is not diagnosed, and nothing here guesses at it.** The suspect is the hardcoded actor `claude-sonnet-4-6` — a `model` the API may no longer serve, which would 400 every request in exactly this pattern — but changing it alters what all nine suites measure, so it waits for the message `d7d157a` makes visible. `EVAL_MODEL` already overrides it if that turns out to be the cause.

This also means **hypothesis 3 is neither confirmed nor falsified**. The retrieval-relevance work in this PR has not been measured yet; `0/14` here is the transport failure, not a score. The gate result in the PR body stands unchanged at `confidence(analysis) = 86%`, provisional.

---

mthines:
## Diagnosed: the Anthropic account is out of credits. Not fixable from this PR.

[Run 34048339749](https://github.com/mthines/agent-skills/actions/runs/34048339749/job/101527328699) on `d7d157a` printed the message the previous run was truncating:

```text
API 400: {"type":"error","error":{"type":"invalid_request_error",
"message":"Your credit balance is too low to access the Anthropic API.
Please go to Plans & Billing to upgrade or purchase credits."}}
```

`149/149 cases never reached the model`, and the new banner classified it as `INVALID RUN` rather than a gate failure — which is what it is.

**This is a repository-secret / billing state, not a code state.** A credit balance cannot be caused by a diff, is identical on every branch, and there is no fix to port. The `workflow_dispatch` on `main` I queued as a base-branch check is redundant now: the message settles it more directly than a second red would.

**What unblocks it:** credits on the account behind the `ANTHROPIC_API_KEY` repository secret. That is the one thing here I cannot do. Once they exist, `l2` needs no code change — re-run the check, or dispatch the single suite (`workflow_dispatch` → `suite: code-review-retrieval-relevance`, ~30 s rather than ~6 min for all nine).

**Two corrections to my previous comment**, both mine:

- I wrote **189 cases**. The correct total is **149** (30+18+14+15+15+6+15+22+14). The run's own banner says 149; I had summed it wrong.
- I named `claude-sonnet-4-6` as the suspect, reasoning that a retired model ID would 400 every request in this pattern. It would — but that is not what happened. The actor is fine and was correctly left alone; a guess-and-change there would have altered what all nine suites measure while fixing nothing.

**One defect in my own fix, which the run exposed** (`a494a6c`). The dedup `Map` from `d7d157a` keyed on the raw error message, and an Anthropic error body ends in a unique `"request_id":"req_…"` — so it deduplicated nothing and printed 149 separate `1×` lines for one cause. Dedup that does not dedup is worse than none: the same wall of noise, now claiming to be a summary. The key now strips the per-request identifier, covering both shapes the API returns (`"req_…"` and `null`). Verified against both real bodies from this run and the local invalid-key self-test: five errors of two causes collapse to two entries with counts 3 and 2; the live self-test reports `14× API 401 …` on one line instead of fourteen. L1 stays 1519/1519.

**Status of the actual work in this PR is unchanged and unmeasured.** Hypothesis 3 has still not been tested — `0/14` was transport, not a score — so `confidence(analysis) = 86%` stands, provisional, and the eval-definition half of this PR is not a confirmed fix. The three rubric fixes (the `source.agent` self-contradiction, the exact-match `scope` statement, and the `surfaces` verb) do not depend on the eval running; they repair the thing under test.

Keeping the PR watched. Nothing further to push until the credit balance changes.

---

PR #185, opened 2026-09-07
Make the detection eval measurable before tuning what it measures

mthines:
## What

Three changes to `bug-detection`, all to the **measuring instrument**. Neither rubric is edited — `finders.md` and `finding-verifier.md` are untouched, and so is the gate.

1. **Cached system prefixes + a token line** in `l2-detection.mjs`.
2. **Controls grown 10 → 30**, so `fp_rate` can express a verdict against its own gate.
3. **Five defective controls repaired**, after the new diagnostic named them.

Plus the run records for 6–10 in `scripts/eval/README.md` and `CLAUDE.md`, and one new reference file drawing the general lessons out of the episode (see [Also in this PR](#also-in-this-pr)).

## Why

The detection baseline's own prescription is *repeat runs before and after any rubric edit, three as a floor*. Two things made that prescription unaffordable and, worse, unreadable.

**The runner was uninstrumented.** Both system blocks are whole rule files (~2.5k tokens each) and were re-sent uncached on every call — the verifier's once per *candidate*, not per record — and the run printed nothing about what it cost. `l2.mjs` fixed exactly this months ago. Measured effect: **~74% lower input cost**, stable across all five runs since.

**`fp_rate` could not resolve the difference it was being asked about.** At 10 controls one control was 10 points against a 20-point gate, so the metric had four usable states: 0, 10, 20, 30, 40. The five recorded runs land exactly on that lattice — 30 · 40 · 30 · ≤20 · 40. A rubric edit moving precision from 27% to 21% would have read as *no change*. Indistinguishable from the run-to-run variance runs 4 and 5 demonstrate on identical content.

## The controls

20 new records, four per class family. **Every one is a decoy** — a boring diff would measure diff difficulty rather than discrimination — and each probes a named finder failure mode:

| family | what the four probe |
| --- | --- |
| `logic` | the defect removed; code that IS the fix for a classic bug (`((i % n) + n) % n`); integer cents where a float rule does not apply; a deliberately serial `await` loop |
| `consumer-break` | an *optional* field added; a parameter type *widened*; a rename with a deprecated alias in the same diff; an added overload |
| `dep-breaking-change` | a purely additive minor; a `0.x` **patch** (zerover promotes a minor, not a patch); a real major whose usage sites the same diff updates; a devDependency major hitting test files only |
| `intent-mismatch` | a refactor honestly described; a perf claim with its benchmark in the diff; extra scope *disclosed*; a revert described as a revert |
| `standards` | raw SQL in the layer the rule permits; the logger *replacing* console; a rollback in the paired `down.sql` the convention specifies; `any` in a test file the rule exempts |

## What the measurement then found — and why the verifier was NOT tuned

The plan after growing the controls was to tune `finding-verifier.md` for precision. **It was not done, and should not be.** Run 7 was the first to print its surviving controls *by name*, and four and a half of the five false positives were **defective fixtures** — the finder was correct each time (a `shallowClone` spreading primitives, a `Future` used with no import, a "clean" perf diff introducing module-level state and an unenforced call ordering, a `mockReset()` break the record wrongly excused as "only a devDependency").

Tuning the verifier against those would have trained it to reject **true** findings while `fp_rate` went down — the same fix-to-pass the gate exists to prevent, one level up and much harder to notice. So the fixtures were repaired instead. Two of the five were **original** controls whose defects had been invisible for the whole recorded baseline.

Two rules came out of it, both recorded:

- **A control asserting "this diff is clean" is a claim about the whole diff, not about the one thing it probes.** Four of five probed correctly and were dirty for an unrelated reason.
- **`fp_rate` alone cannot distinguish an over-permissive verifier from a defective control set.** Five runs of a bare number pointed at the rubric; one run of named survivors pointed at the fixtures.

## Measured runs

| run | recall | fp_rate | lift | gate |
| --- | --- | --- | --- | --- |
| 6 (pre-repair) | 90% | 20% (6/30) — on the boundary | +13 | pass |
| 7 (pre-repair) | 80% | 17% (5/30) | +20 | pass |
| 8 (post-repair) | 80% | **10%** (3/30) | +12 | pass |
| 9 (repeat) | 80% | **10%** (3/30) | +5 | pass |
| 10 (repeat) | **70%** — *at the gate* | **7%** (2/30) | +18 | pass |

Runs 9 and 10 are **true repeats** — their commits change only prose, so fixtures, both rubrics and the runner are byte-identical.

Two findings that only the repeats could produce:

- **Run 9 matched run 8's headline while a third of the set behind it moved.** Two survivors stable, two swapped across the verifier boundary — including one the run-7 write-up had cited as evidence *for* the core, which turned out to be a sixth over-loaded fixture. A bare matching rate would have read as a settled operating point.
- **Run 10 moved recall 10 points on identical inputs, landing exactly on the gate.** That is the *same defect the controls just had*, on the other half of the set: at 20 seeded records one case is 5 recall points, so a rubric edit worth under ~10 points is indistinguishable from sampling.

## The next piece of work changed as a result

**Grow the seeded half**, which now outranks the `intent-mismatch` work. Same argument as the controls: never tune a rubric against a number that cannot express the change. `fp_rate` is now measurable and falling (10 → 10 → 7) with only two known-borderline originals surviving; recall is the half that is not yet measurable.

## Guards, each probed by breaking it

- `floor(controlN × GATES.fp) >= 4` — reverting to 10 controls reds it. **The fix for a breach is always more controls**: lowering `GATES.fp` shrinks the tolerated count, degrading resolution while looking like a concession.
- Both system prefixes clear the 1024-token minimum — so unlike `l2.mjs` (which has three genuinely-small suites), a `cache MISSED` here is a real defect. Raising the bound above the real prefix reds it.
- The cache saving is positive — inverting the read/write cost weights turns it negative and reds it.

`node scripts/eval/l1.mjs` → 1612/1612. `--self-test` green.

## Also in this PR

`agents/shared/references/self-improvement-evidence.md`, linked from `memory-buckets.md`'s See also.

`memory-buckets.md` says what the LoreKit buckets are; nothing said why they are worth the wiring, and no skill here could point at a case where a missing carried lesson cost something countable. This episode can, so the general lessons are drawn out where the next author will find them rather than left buried in ten run write-ups.

It states its own limits first: **this episode did not run on LoreKit** — the lessons were hand-written into the slow tier — so it is not a demonstration that the store delivered anything. What makes it evidential is an internal control: four general lessons that were *not* written down were each re-derived 1–4 times in one session, and the one that *was* already written down was re-derived zero times and held at run 10, where recall landed on the gate and a two-point move would have bought margin. It also records the failure mode on the other side — two of the five lessons were wrong when first written, so a carried lesson is a carried belief.

## Accepted costs, stated

- A full run is 50 records rather than 30, and `--class` keeps all controls by design (a per-class recall figure with no false-positive rate beside it is the half that can be gamed), so even a single-class run pays the 30. The new token line is what makes that visible; the cached prefixes are what make it affordable.
- **Every `fp_rate` in the pre-growth baseline is over a 10-control denominator and is not comparable as a number to a current run's.** The README and `CLAUDE.md` both say so at the point of use, and dirty-control counts are quoted alongside each rate so the underlying observation stays legible.
- Two fixture repairs are owed and named, not done here: `control-description-matches-larger-diff`, and a consumer surface for `control-off-by-one-that-is-correct`.

## Evals

Per [the maintenance table](https://github.com/mthines/agent-skills/blob/main/CLAUDE.md#keeping-the-evals-honest-mandatory-on-every-change): this changes `l2-detection.mjs` and `golden/bug-detection.jsonl`, which select **`bug-detection` and no L2 suite** (verified with `select-suites.mjs`). Labelled `run-evals`; five measured runs are reported above.

## Also

Corrects claims in `CLAUDE.md` that later runs had already falsified — the sample size, and the "precision gap is the reproducible failure" reading, which is a flaky gate rather than a standing red.

---
https://claude.ai/code/session_01EFoqtKWwjJcRFDxQ1Aowss

---
_Generated by [Claude Code](https://claude.ai/code/session_01EFoqtKWwjJcRFDxQ1Aowss)_


mthines:
### 🟡 4 findings

Instruments and widens the `bug-detection` eval without touching either rubric it measures, the gate, or any seeded record.

**Warnings:** 4 non-blocking findings

|---|---|---|
| Point-free map changes the call arity | `scripts/eval/golden/bug-detection.jsonl:43` | 🟡 medium |
| Memoised lookup goes stale if entries is mutated | `scripts/eval/golden/bug-detection.jsonl:44` | 🟡 medium |
| CREATE INDEX without CONCURRENTLY in a control | `scripts/eval/golden/bug-detection.jsonl:49` | ⚪ low |
| Control cleanliness rests on an unshown usage site | `scripts/eval/golden/bug-detection.jsonl:42` | ⚪ low |

<details>

**Needs attention**

| Gate | Status | Details |
|---|---|---|
| Description vs. code | ✅ | Every load-bearing claim in the description was checked against the diff and holds. |
| Documentation | ✅ | README gains 302 lines of run records; CLAUDE.md corrections match the measurements. |
| Self-review signals | ✅ | No debug logs, leftover markers, suppressions, or secrets on new lines. |
| Code review | ⚠️ | 4 non-blocking findings, all on new control fixtures — see inline comments. |

<details>
<summary>Withheld (1) — could not be verified from this runner</summary>

- `scripts/eval/README.md:943` — question: The five reported run figures could not be reproduced here. <sup>(unverified: the five reported run figures need a live ANTHROPIC_API_KEY to reproduce; unverifiable offline)</sup>

</details>

**Found**

Quality — produced 4 → posted inline 4 · cleared 4 · carried forward 0 · deferred 0 · below-bar 0
Severity — 🟡 2 medium · ⚪ 2 low
Verified — Re-ran offline: `l1.mjs` 1612/1612 · `l2-detection.mjs --self-test` green · `select-suites.mjs` → `bug-detection` only. `GATES` byte-identical at base and head; `finders.md` and `finding-verifier.md` absent from the change set; all 20 seeded records byte-identical; no record removed or reclassified. All three new guards were probed by breaking them and each went red.

**Run**

full · 542 lines in delta · tier deep · depth checkout · 4 files · first review of this PR · no prior state record
⚠️ Tool-call budget exceeded (~36 of 30 for a 4-file diff) — spent on offline verification of the PR's own executable claims, not on extra scanning. Coverage is complete.
CI — `l1`, `opt-in gate`, `select affected suites` green; `suite` correctly skipped (no L2 suite selected); `bug-detection` still in progress.
Memories — 0 indexed · 0 used

<sup>Nothing to report — standards (1 doc), optimality (1 judged), measurability (skipped), integrations (not activated), 0 files skipped.</sup>

</details>

<sup>`pr-reviewer` · commit `c576feb` · full review · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md) · updated <relative-time datetime="2026-09-08T06:00:00.000Z">Sep 8, 2026 6:00am UTC</relative-time></sup>

---

scripts/eval/golden/bug-detection.jsonl
```diff
@@ -1,30 +1,50 @@
...
+{"id":"control-refactor-described-as-refactor","class":"control","diff":"--- a/src/lib/parse.ts\n+++ b/src/lib/parse.ts\n@@ -6,12 +6,12 @@ import type { Token } from \"./types\";\n\n export function parse(src: string): Token[] {\n-  const out: Token[] = [];\n-  for (const line of src.split(\"\\n\")) {\n-    if (!line.trim()) continue;\n-    out.push(tokenize(line));\n-  }\n-  return out;\n+  return src\n+    .split(\"\\n\")\n+    .filter((line) => line.trim())\n+    .map(tokenize);\n }","defect":null,"notes":"The description claims a behaviour-preserving refactor and the diff is exactly that: same filter, same order, same output. The decoy for an intent finder that reports a mismatch whenever the diff is shaped differently from the prose describing it.","description":"Rewrite parse() in a functional style\n\nNo behaviour change: same blank-line filter, same tokenizer, same output order.\nPurely a readability pass ahead of the streaming work."}
```
mthines, line 43: suggestion (medium): 🟡 **Point-free map changes the call arity** **(non-blocking)**

The original passed one argument; `.map(tokenize)` passes (element, index, array), so "same output" holds only if `tokenize` is unary — and its signature is on no surface the finder is shown.

Evidence: `scripts/eval/golden/bug-detection.jsonl:43` (notes claim same output)

```diff
-    out.push(tokenize(line));
+    .map(tokenize);   // now tokenize(line, i, arr)
```

<sup>`pr-reviewer` · commit `c576feb`</sup>



---

scripts/eval/golden/bug-detection.jsonl
```diff
@@ -1,30 +1,50 @@
...
...ms/op\n+// after:  0.0009 ms/op (steady state; the first call per array builds the index)","defect":null,"notes":"The intent-perf-claim-unsupported record with the evidence supplied: the diff carries the benchmark the description cites, the exported signature is unchanged, and the memo is keyed on the caller's own array so there is no shared state and no call-ordering rule. The decoy for an intent finder that flags every performance claim as unsupported without checking whether the diff supports it. (Run 7 note: the first version of this record used a module-level Map populated by a separate exported `index()`, and all five surviving candidates were CORRECT about it — unenforced ordering, cross-caller state, a silent `undefined` before the first index(). The perf claim was honest; the code carrying it was not clean, which made this a defective control rather than a decoy.)","description":"Index lookups by key instead of scanning\n\nlookup() was a linear scan on every call. It now builds a Map on first use and memoises\nit against the caller's own entries array, so the exported signature is unchanged.\nBenchmark in the same commit: 1.42 ms/op → 0.0009 ms/op on a 50k-entry fixture."}
```
mthines, line 44: suggestion (medium): 🟡 **Memoised lookup goes stale if entries is mutated** **(non-blocking)**

The WeakMap is keyed on array identity, so a push after the first lookup is never seen — a behaviour change from `entries.find()` the note's no-shared-state argument does not cover.

Evidence: `scripts/eval/golden/bug-detection.jsonl:44` (repaired once for this)

```ts
let byKey = indexes.get(entries);
if (!byKey) { /* built once; later mutations to entries are invisible */ }
```

<sup>`pr-reviewer` · commit `c576feb`</sup>



---

scripts/eval/golden/bug-detection.jsonl
```diff
@@ -1,30 +1,50 @@
...
+{"id":"control-migration-rollback-in-down-file","class":"control","diff":"--- /dev/null\n+++ b/migrations/00121_add_tax_region/up.sql\n@@ -0,0 +1,2 @@\n+alter table invoices add column tax_region text;\n+create index invoices_tax_region_idx on invoices (tax_region);\n--- /dev/null\n+++ b/migrations/00121_add_tax_region/down.sql\n@@ -0,0 +1,2 @@\n+drop index invoices_tax_region_idx;\n+alter table invoices drop column tax_region;","defect":null,"notes":"The rollback exists, in the paired-file convention the governing doc actually specifies. The decoy for a standards finder that has internalised \"a migration needs a rollback\" as \"the migration file needs a rollback comment in it\" and reports the up.sql in isolation.","governing_doc":{"path":"docs/migrations.md","excerpt":"## Every migration is reversible\n\nEach migration is a directory containing `up.sql` and `down.sql`. A migration without a\n`down.sql` is rejected by CI. Do not put rollback statements inside `up.sql`."}}
```
mthines, line 49: question (low): `create index` here has no `concurrently`, which the correctness finder’s own schema-migration checklist names — unrelated to the rollback-placement probe. Intended? **(non-blocking)**

<sup>`pr-reviewer` · commit `c576feb`</sup>

---

scripts/eval/golden/bug-detection.jsonl
```diff
@@ -1,30 +1,50 @@
...
..."class":"control","diff":"--- a/package.json\n+++ b/package.json\n@@ -34,7 +34,7 @@\n   \"devDependencies\": {\n-    \"vitest\": \"^2.1.8\",\n+    \"vitest\": \"^3.0.4\",","defect":null,"notes":"A real major with real breaking changes, neither of which this repo's one usage site touches: the changelog changes `mockReset()`, the test calls `mockClear()`, and the pool default is not configured here. The decoy for a dependency finder that matches an API FAMILY rather than the exact symbol the changelog names. (Run 7 note: this record previously pointed at `mockReset()` and asserted the break was harmless because it landed in a test file — which is false, a broken test is broken, and the finder was right to report it.)","dependency":{"name":"vitest","from":"2.1.8","to":"3.0.4","delta":"major","changelog":"## 3.0.0\n### Breaking\n- The default `pool` is now `forks`; `threads` must be requested explicitly.\n- `spy.mockReset()` restores the original implementation rather than leaving an empty mock.","usage_sites":[{"path":"src/lib/format.test.ts","line":21,"excerpt":"20:   const spy = vi.spyOn(clock, \"now\");\n21:   spy.mockClear();\n22:   expect(clock.now()).toBeTypeOf(\"number\");"}]}}
```
mthines, line 42: question (low): The note asserts the one usage site calls `mockClear()`, but no test surface is in the diff — the same gap the PR names as owed for control-off-by-one-that-is-correct. **(non-blocking)**

<sup>`pr-reviewer` · commit `c576feb`</sup>

---

mthines (commented):
---

PR #186, opened 2026-09-09
create-skill 1.4.0: align with the current skill spec, add a mechanical validator and an eval phase

mthines:
## Why

A review of `create-skill` against the current Claude Code skill docs, the Agent Skills open spec, and Anthropic's authoring best practices found it structurally sound but stale in five areas: the frontmatter reference lacked `disallowed-tools`, `background`, `compatibility`, and the six-field portability profile; nothing covered string substitutions or dynamic `!`cmd`` injection; there was no evaluation phase; the skill violated its own TOC rule on nine of its files; and review mode was pure judgment where roughly half the checklist is mechanically checkable.

## What changed

- **Four new rules** — `arguments-and-injection.md`, `evaluation.md`, `scripts-and-assets.md`, `workflow-patterns.md`. Ten existing rules updated: tightened `name` regex, portability profile and the hard error it prevents, skill locations and precedence, the 1 % listing budget with `/skill-doctor`, trigger eval sets, five new anti-patterns.
- **Scaffold gained Phase 6 (Evaluation)** plus two Phase 0 questions (target runtime, objectively verifiable output). New templates: `SKILL.portable.md`, `evals.json`, `triggers.jsonl`.
- **`scripts/validate-skill.mjs`** — zero-dependency mechanical pre-pass, 31 check rows across 24 finding ids, with `--portable` / `--json` / `--strict` and a 27-assertion `--self-test`. Review mode and the Phase 5 self-check run it first; every checklist item is now tagged `(mechanical: <id>)` or `(judgment)`.
- **L1 `G51`** — runs the self-test, runs the validator on the skill itself, asserts the script cannot be orphaned, enforces the TOC rule on the skill's own long files, and (`G51e`) pins the `${CLAUDE_SKILL_DIR}` path convention plus SC01's discrimination behaviourally. Every sub-check was proven to bite by breaking it and observing the named row go red.
- "Used by" layout examples re-verified against the tree; `create-pr` was cited as single-file but is a 515-line multi-file skill.

## Two profile-split decisions worth flagging

Both came from primary sources rather than from inference, because the obvious reading of each rule is wrong:

- **`FM07` (angle brackets).** The Skills API upload path rejects *any* `<` or `>` in a description (Anthropic's own `quick_validate.py`), but `frontmatter.md` mandates `<…>` for CLI argument placeholders, and Claude Code itself has no such rule. So the check is profile-aware: a real XML tag fails everywhere, a bare placeholder fails only under `--portable`.
- **`FM03` (reserved words).** The Skills API states a *substring* rule for `anthropic` / `claude`; Claude Code's docs carry no reserved-word rule at all, only the reserved `synced` folder name. So `synced` is a case-insensitive hard FAIL, and the substring rule is a `WARN` that becomes a `FAIL` only under `--portable` — otherwise the shipped `optimize-claude-md` would hard-fail on its own name.

## One known coverage boundary, stated deliberately

`SC01` and `G51e` both enforce the `${CLAUDE_SKILL_DIR}` path convention by matching an interpreter prefix, and that interpreter list is an allowlist which **cannot be made complete** — `ruby scripts/x.sh` is not caught. This is recorded in the code rather than papered over with an assertion that could not establish completeness. It is the safe direction: a missing interpreter is a missed violation, never a false accusation, and the false-positive side would land on other people's skills while this false-negative side is confined to `create-skill`'s own docs.

## Evals

No L2 suite reads any `create-skill` file (`select-suites.mjs` selects none), so the new mechanical contract is guarded by L1 rather than by a behavioural suite. L1: **1638/1638**.

Informational: the validator's sweep over all 55 skills in the repo shows **24 clean and 31 failing**, mostly bare code fences and missing TOCs. Nothing outside `create-skill` was touched here, and no corpus count moved across any of this PR's five review iterations.

https://claude.ai/code/session_01BVd9pb1wz6RbRP3TCqPCvS

---
_Generated by [Claude Code](https://claude.ai/code/session_01BVd9pb1wz6RbRP3TCqPCvS)_


skills/authoring/create-skill/SKILL.md
```diff
@@ -195,35 +212,37 @@ If the user is publishing the skill via `npx skills add` only, skip steps
...
 
-Run the full checklist in `rules/quality-checklist.md`. Treat any
-unchecked item as a defect — fix it before declaring the skill done.
-
-Report the result inline as:
-
-```
-Self-check: PASS (28/28)
-```
+Run `node skills/authoring/create-skill/scripts/validate-skill.mjs <dir> [--portable]`
```
mthines: issue (high): 🟠 **Documented validator command cannot run outside this repo** **(blocking)**

Invoked by a cwd-relative repo path at four sites, so it throws MODULE_NOT_FOUND once the skill is installed by symlink — the form this PR's own new rule forbids.

Evidence: `skills/authoring/create-skill/rules/scripts-and-assets.md:59` (the rule this violates) · `skills/authoring/create-skill/SKILL.md:241`

```bash
node ${CLAUDE_SKILL_DIR}/scripts/validate-skill.mjs <dir> [--portable]
```

<sup>`pr-reviewer` · commit `e5b3a39`</sup>



---

skills/authoring/create-skill/scripts/validate-skill.mjs
```diff
@@ -0,0 +1,704 @@
...
+      "lens.md", null,
+      `lens.md must be <= ${LENS_MAX} lines and declare for: pr-reviewer + lens-version: 1 (lines=${lc}, for=${JSON.stringify(forVal)}, lens-version=${JSON.stringify(lensVersion)})`);
+  } else {
+    f.check("PD04", "WARN", true, "SKILL.md", null, "");
+  }
+
+  // SC01: scripts/ existing implies SKILL.md should reference it.
+  const scriptsDir = join(dir, "scripts");
+  if (existsSync(scriptsDir) && statSync(scriptsDir).isDirectory()) {
+    f.check("SC01", "WARN", text.includes("${CLAUDE_SKILL_DIR}") || /\bscripts\//.test(text), "SKILL.md", null,
```
mthines: issue (medium): 🟡 **SC01's OR arm is satisfied by the form it forbids** **(non-blocking)**

A cwd-relative `scripts/...` path matches the second arm, so SC01 passes on the very violation it exists to catch — which is why create-skill self-validates clean.

Evidence: `skills/authoring/create-skill/rules/scripts-and-assets.md:59` (convention)

```js
f.check("SC01", "WARN", text.includes("${CLAUDE_SKILL_DIR}"), "SKILL.md", null,
```

<sup>`pr-reviewer` · commit `e5b3a39`</sup>



---

skills/authoring/create-skill/scripts/validate-skill.mjs
```diff
@@ -0,0 +1,704 @@
...
+    "SKILL.md", descLine,
+    `description must be non-empty and <= ${DESC_MAX} chars, got ${description ? description.length : 0}`);
+
+  const whenToUse = getText(map, "when_to_use");
+  const combinedLen = (description ? description.length : 0) + (whenToUse ? whenToUse.length : 0);
+  f.check("FM06", "WARN", combinedLen <= DESC_WHEN_MAX, "SKILL.md", descLine,
+    `description + when_to_use combined must be <= ${DESC_WHEN_MAX} chars, got ${combinedLen}`);
+
+  const xmlTag = /<[^>]+>/;
+  f.check("FM07", "FAIL",
```
mthines, line 283: issue (medium): 🟡 **FM07 hard-fails CLI argument placeholders** **(non-blocking)**

The bare `<[^>]+>` regex reads `<fact>`, `<stack>` and `<scope>` as XML tags, FAILing four shipped skills — while frontmatter.md mandates `<…>` for placeholders.

Evidence: `skills/authoring/create-skill/rules/frontmatter.md:34` (mandates the form) · `skills/quality/pr-review/SKILL.md:3` (remember <fact>)

```js
const xmlTag = /<\/?[a-zA-Z][a-zA-Z0-9-]*(\s[^>]*)?>/;
```

<sup>`pr-reviewer` · commit `e5b3a39`</sup>



---

skills/authoring/create-skill/scripts/validate-skill.mjs
```diff
@@ -0,0 +1,704 @@
...
+  const name = getText(map, "name");
+  const nameLine = getLine(map, "name");
+
+  f.check("FM02", "FAIL",
+    !!name && name.length <= NAME_MAX && /^[a-z0-9]+(-[a-z0-9]+)*$/.test(name),
+    "SKILL.md", nameLine,
+    `name must match ^[a-z0-9]+(-[a-z0-9]+)*$ and be <= ${NAME_MAX} chars, got ${JSON.stringify(name)}`);
+
+  const nameIsClean = !!name && !/anthropic|claude/i.test(name) && dirName !== "synced";
+  f.check("FM03", "FAIL", nameIsClean, "SKILL.md", nameLine,
```
mthines: issue (medium): 🟡 **FM03 substring test fails a shipped skill name** **(non-blocking)**

`optimize-claude-md` contains "claude" and hard-FAILs, but the rule it cites forbids reserved words (`claude-tools`, `anthropic-helper`), not the substring anywhere.

Evidence: `skills/authoring/create-skill/rules/frontmatter.md:165` (reserved words) · `skills/authoring/optimize-claude-md/SKILL.md:2` (shipped skill)

<sup>`pr-reviewer` · commit `e5b3a39`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -6335,4 +6335,70 @@ const isPollBlock = (block) =>
...
+  const checklist = existsSync(checklistPath) ? readFileSync(checklistPath, "utf8") : "";
+  s.check("G51c quality-checklist.md names the validator under a Mechanical pre-pass section",
+    /## Mechanical pre-pass/.test(checklist) && /validate-skill\.mjs/.test(checklist),
+    existsSync(checklistPath) ? "no ## Mechanical pre-pass section naming validate-skill.mjs" : "rules/quality-checklist.md not found");
+
+  // (d) The skill obeys its own TOC rule on its own long rule files (rules/*.md > 150 lines
+  // need a ## Contents / ## Table of contents heading — see rules/progressive-disclosure.md).
+  const rulesDir = join(CS, "rules");
+  if (existsSync(rulesDir)) {
+    for (const name of readdirSync(rulesDir).filter((f) => f.endsWith(".md")).sort()) {
```
mthines: suggestion (low): ⚪ **G51d guards only half the TOC rule it cites** **(non-blocking)**

It iterates `rules/` at 150 lines but not `references/` at 100, and PD03 is a WARN so G51b exits 0 either way.

Evidence: `skills/authoring/create-skill/rules/progressive-disclosure.md:149` (both thresholds) · `skills/authoring/create-skill/references/skill-archetypes.md:1` (487 lines, unguarded)

<sup>`pr-reviewer` · commit `e5b3a39`</sup>



---

mthines (commented):
---

mthines:
### 🟡 2 findings

Every one of the five commits lands the iteration-4 finding it claims to, the force-push is tree-identical as described, and all four re-verified counts match; what remains is that the new guard is now stricter than the check it guards.

**Warnings:** two non-blocking findings, both coverage gaps in the guard this delta added rather than regressions in the five fixes

|---|---|---|
| SC01 misses the bare form G51e now catches | [`skills/authoring/create-skill/scripts/validate-skill.mjs:508`](https://github.com/mthines/agent-skills/pull/186#discussion_r3970469744) | 🟡 medium |
| G51e's coverage rests on two hand-kept allowlists | [`scripts/eval/l1.mjs:6452`](https://github.com/mthines/agent-skills/pull/186#discussion_r3970469756) | ⚪ low |

<details>

**Needs attention**

| Gate | Status | Details |
|---|---|---|
| Description vs. code | ✅ | All four re-measured and matching: 31 rows / 24 ids, self-test 26, L1 1638/1638, 24 clean / 31 failing of 55. |
| Prior review feedback | ✅ | All 17 prior threads resolved, 4 since `94818ae`; the rewritten messages' four `discussion_r` ids match real threads. |
| Documentation | ✅ | Each delta hunk carries its own rationale, including why the `ctx.dirName` reuse was a tidy-up and not a live bug. |
| Self-review signals | ✅ | No markers, residue, suppressions, focused tests, or secrets on the 66 added lines of the delta. |
| Code review | ⚠️ | 2 findings survived, neither blocking; both are coverage gaps in the new G51e guard, not regressions in the five fixes. |

<details>
<summary>Impact — 3 changed exports · 1 consumer checked</summary>

- `INTERPRETER` (`skills/authoring/create-skill/scripts/validate-skill.mjs`) — added change · 1 consumer file · 1 verified unaffected
- `checkPortabilityDocs` (`skills/authoring/create-skill/scripts/validate-skill.mjs`) — body change · 0 consumer files
- `selfTest` (`skills/authoring/create-skill/scripts/validate-skill.mjs`) — body change · 7 consumer files · 7 not traced (budget)

</details>

**Found**

Quality — produced 6 → posted inline 2 · cleared 2 · carried forward 0 · deferred 0 · below-bar 0
Dropped — relevance 0 · dedupe 0 · grounding 0 · confidence 0 · shape 0 · contradicted 4
Severity — 🟡 1 medium · ⚪ 1 low
Standards — ran · 1 governing doc compared (`rules/scripts-and-assets.md` § Path convention) · 1 finding
Verified — Executed at `checkout` on the local object store at `77c3687`. The force-push was checked rather than taken on trust: `94818ae` is an ancestor of `77c3687`, so the delta range is a real authored delta and no blob-diff fallback was needed, and all five old→new commit pairs have identical trees. The four description counts were re-measured, not read: L1 1638/1638, validator `--self-test` 26 assertions, 31 `f.check` rows across 24 distinct ids, and 55 skills of which 24 are zero-FAIL and 31 carry at least one FAIL — confirming the orientation the last run's inherited knowledge record had inverted. Both findings carry mutation receipts: adding a `.rb` script plus a `ruby` invocation leaves L1 green at 1638/1638 while the equivalent `.sh` pair reds it at 1637/1638, and the proposed SC01 arm was measured against all 55 skills at 0 new findings before it was proposed — the first remedy considered was refuted the same way, because it fired on four documentation placeholders SC01's own comment protects.

**Run**

incremental · 80 lines in delta · tier standard · depth checkout · 80-line delta under the 100 threshold, no new files, no high-stakes path, no propagation shape, and a deep pass is recorded at `94818ae`; blast_radius=low, semver_delta=none, high_stakes=0, risky_shapes=none, thread_overlap=1.0 (band is not none, so the quick override does not apply)
CI — All green on `77c3687` — `l1` and the `l2` aggregator pass; the three opt-in eval jobs are correctly skipped without the `run-evals` label.
Memories — 56 indexed · 3 used (1 knowledge · 2 hotspots)

- **hotspot** `hotspot::scripts/eval/l1.mjs` — its recorded `test-gap` class, 6 confirmed, is why G51e was mutation-probed rather than read
- **hotspot** `hotspot::skills/authoring/create-skill/scripts/validate-skill.mjs` — its `over-strict-check` half is why the proposed SC01 arm was corpus-measured before it was proposed
- **knowledge** `knowledge::validateSkill@skills/authoring/create-skill/scripts/validate-skill.mjs` — its 24 check ids re-verified at 24; its corpus orientation was inverted and is corrected by this run

<sup>Nothing to report — optimality (skipped), measurability (skipped), integrations (not activated), 0 files skipped.</sup>

</details>

<sup>`pr-reviewer` · commit `77c3687` · incremental review, delta since `94818ae` · [how these findings are produced](https://github.com/mthines/agent-skills/blob/main/agents/pr-reviewer.md) · updated <relative-time datetime="2026-09-09T16:05:00.000Z">Sep 9, 2026 4:05pm UTC</relative-time></sup>


---

skills/authoring/create-skill/SKILL.md
mthines: Addressed in 2b25575. ✅

---

skills/authoring/create-skill/scripts/validate-skill.mjs
mthines: Addressed in 03a4457. ✅

---

skills/authoring/create-skill/scripts/validate-skill.mjs
mthines: Addressed in 7d49c15. ✅

---

skills/authoring/create-skill/scripts/validate-skill.mjs
mthines: Addressed in 60358e9. ✅

---

scripts/eval/l1.mjs
mthines: Addressed in f9aa68f. ✅

---

skills/authoring/create-skill/scripts/validate-skill.mjs
```diff
@@ -0,0 +1,785 @@
...
+    f.check("PD04", "WARN", true, "SKILL.md", null, "");
+  }
+
+  // SC01: scripts/ existing implies SKILL.md invokes at least one script through
+  // ${CLAUDE_SKILL_DIR} (rules/scripts-and-assets.md § Path convention). A bare
+  // `scripts/` mention is NOT accepted — a cwd-relative path is exactly the form the
+  // rule forbids, and accepting it let create-skill self-validate clean on a violation.
+  const scriptsDir = join(dir, "scripts");
+  if (existsSync(scriptsDir) && statSync(scriptsDir).isDirectory()) {
+    f.check("SC01", "WARN", text.includes("${CLAUDE_SKILL_DIR}"), "SKILL.md", null,
```
mthines: issue (medium): 🟡 **SC01 passes on a prose mention of the convention** **(non-blocking)**

The narrowed predicate is still a whole-file `text.includes`, so a SKILL.md that only describes the convention passes while every invocation stays cwd-relative.

Evidence: `skills/authoring/create-skill/SKILL.md:180` (the satisfying prose mention) · `skills/authoring/create-skill/rules/scripts-and-assets.md:59` (the convention)

```js
f.check("SC01", "WARN", /\$\{CLAUDE_SKILL_DIR\}\/scripts\//.test(text), "SKILL.md", null,
```

<sup>`pr-reviewer` · commit `8325d4f`</sup>



---

skills/authoring/create-skill/SKILL.md
```diff
@@ -195,35 +212,37 @@ If the user is publishing the skill via `npx skills add` only, skip steps
...
-
-Report the result inline as:
-
-```
-Self-check: PASS (28/28)
-```
+Run `node ${CLAUDE_SKILL_DIR}/scripts/validate-skill.mjs <dir> [--portable]`
+first, then work through the remaining `(judgment)` items in
+`rules/quality-checklist.md`. Treat any unchecked item as a defect — fix it
+before declaring the skill done. Report inline: `Self-check: PASS (28/28)`,
```
mthines: issue (medium): 🟡 **Phase 5 prescribes a self-check count nothing emits** **(non-blocking)**

The phase now runs the validator first, which prints its own `Self-check: PASS (70/73)`. The literal `28/28` matches neither that nor the checklist's 57 items.

Evidence: `skills/authoring/create-skill/scripts/validate-skill.mjs:513` (prints Self-check: PASS (pass/total)) · `skills/authoring/create-skill/rules/quality-checklist.md:12` (57 checklist items)

```text
before declaring the skill done. Report the validator's own `Self-check:` line verbatim,
```

<sup>`pr-reviewer` · commit `8325d4f`</sup>



---

skills/authoring/create-skill/scripts/validate-skill.mjs
```diff
@@ -0,0 +1,785 @@
...
+    const warnRow = withoutFlag.findings.find((x) => x.id === "FM03" && x.level === "WARN");
+    t("reserved word without --portable: FM03 is a WARN", !!warnRow, JSON.stringify(withoutFlag.findings.filter((x) => x.id === "FM03")));
+    t("reserved word without --portable: the skill still passes", withoutFlag.ok === true);
+    const withFlag = validateSkill(dir, { portable: true });
+    const failRow = withFlag.findings.find((x) => x.id === "FM03" && x.level === "FAIL");
+    t("reserved word with --portable: FM03 is a FAIL", !!failRow, JSON.stringify(withFlag.findings.filter((x) => x.id === "FM03")));
+  });
+
+  const ok = fails === 0;
+  console.log(ok ? "validate-skill self-test: PASS (13 assertions)" : `validate-skill self-test: FAIL (${fails} failure(s))`);
```
mthines: issue (low): ⚪ **Self-test banner undercounts its own assertions** **(non-blocking)**

The three new blocks added eight assertions; the banner moved 10 to 13 while the real count went 14 to 22. It is the only coverage signal the self-test emits.

Evidence: `skills/authoring/create-skill/scripts/validate-skill.mjs:574` (22 call sites, instrumented)

```js
console.log(ok ? `validate-skill self-test: PASS (${total} assertions)` : ...
```

<sup>`pr-reviewer` · commit `8325d4f`</sup>



---

scripts/eval/l1.mjs
```diff
@@ -6335,4 +6335,70 @@ const isPollBlock = (block) =>
...
+  const checklistPath = join(CS, "rules/quality-checklist.md");
+  const checklist = existsSync(checklistPath) ? readFileSync(checklistPath, "utf8") : "";
+  s.check("G51c quality-checklist.md names the validator under a Mechanical pre-pass section",
+    /## Mechanical pre-pass/.test(checklist) && /validate-skill\.mjs/.test(checklist),
+    existsSync(checklistPath) ? "no ## Mechanical pre-pass section naming validate-skill.mjs" : "rules/quality-checklist.md not found");
+
+  // (d) The skill obeys its own two-branch TOC rule (rules/progressive-disclosure.md § Long
+  // reference and rule files need a TOC): rules/*.md past 150 lines and references/*.md past
+  // 100 lines both need a ## Contents / ## Table of contents heading. Both branches are guarded
+  // here because the validator grades PD03 as WARN, so G51b exits 0 either way.
```
mthines, line 6390: suggestion (medium): 🟡 **G51 guards the TOC WARN but not the path WARN** **(non-blocking)**

This comment's own reasoning — a WARN leaves G51b at exit 0 — applies to SC01 too, whose violation was this PR's blocking finding, and no sub-check covers it.

Evidence: `scripts/eval/l1.mjs:6369` (G51b reads run.status only) · `skills/authoring/create-skill/scripts/validate-skill.mjs:471` (SC01 is graded WARN)

```js
s.check("G51e SC01 fires on a cwd-relative script path", scOnFixture === "fires");
```

<sup>`pr-reviewer` · commit `8325d4f`</sup>



---

mthines (commented):
Reviewed `8325d4f` — 4 findings — none blocking; 2 gates warn. 4 finding(s) inline. Sticky exists but this access path cannot edit an issue comment — report not updated in place.

---

skills/authoring/create-skill/scripts/validate-skill.mjs
mthines: Addressed in bf4ebd2. ✅

---

skills/authoring/create-skill/SKILL.md
mthines: Addressed in 6aadde1. ✅

---

skills/authoring/create-skill/scripts/validate-skill.mjs
mthines: Addressed in b30d4e3. ✅

---

scripts/eval/l1.mjs
mthines: Addressed in 23ed485. ✅

---

scripts/eval/l1.mjs
```diff
@@ -6335,4 +6335,104 @@ const isPollBlock = (block) =>
...
+  //      self-test, so this check pins the fixture's existence and G51a proves it passes;
+  //      together they mean the guard cannot go green against a check that stopped firing.
+  //      Asserted against the self-test rather than a temp fixture built here, so the coverage
+  //      lives with the validator that owns SC01 and l1.mjs writes nothing to disk.
+  s.check("G51e SKILL.md invokes the validator through a ${CLAUDE_SKILL_DIR}/scripts/ path",
+    /\$\{CLAUDE_SKILL_DIR\}\/scripts\/validate-skill\.mjs/.test(skillMd),
+    existsSync(skillMdPath)
+      ? "no ${CLAUDE_SKILL_DIR}/scripts/validate-skill.mjs invocation in SKILL.md"
+      : "SKILL.md not found");
+  const cwdRelative = skillMd.match(/node\s+skills\/authoring\/create-skill\/scripts\/validate-skill\.mjs/g) ?? [];
```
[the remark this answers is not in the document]
mthines: Addressed in 368c28c, refined in 94818ae. ✅

Confirmed your receipt first: reverting `rules/quality-checklist.md:41` did leave L1 green at 1634/1634. The arm now walks every markdown file the skill ships and reports each hit with its path, so the detail line points at the offending site rather than at the group.

Mutation-proven on exactly the site you named — the revert now reds L1 at 1635/1636 with `skills/authoring/create-skill/rules/quality-checklist.md: node skills/authoring/create-skill/scripts/validate-skill.mjs`, and the restore returns green.

94818ae is a follow-up simplify pass: the first version hand-rolled a recursive walk that `l1.mjs` already imports as `walk` from `lib.mjs`, and named its loop variable `rel`, shadowing the imported `rel` helper. Both now use the existing helpers, which also upgraded the detail line to a full repo-relative path.

---

skills/authoring/create-skill/scripts/validate-skill.mjs
```diff
@@ -0,0 +1,810 @@
...
+  // rule forbids, and accepting it let create-skill self-validate clean on a violation.
+  //
+  // Scope: the check demands an INVOCATION PATH (`${CLAUDE_SKILL_DIR}/scripts/`), not a
+  // mention of the variable. A prose mention of the convention is not an invocation — it
+  // documents the rule while every actual command in the file stays cwd-relative, which is
+  // precisely the violation this check exists to catch, so a whole-file scan for the bare
+  // variable satisfied SC01 on the very state it is meant to flag.
+  const scriptsDir = join(dir, "scripts");
+  if (existsSync(scriptsDir) && statSync(scriptsDir).isDirectory()) {
+    f.check("SC01", "WARN", /\$\{CLAUDE_SKILL_DIR\}\/scripts\//.test(text), "SKILL.md", null,
```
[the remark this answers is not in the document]
mthines: Addressed in 389d1eb. ✅

The diagnosis is right, and it is the second time this check has been fixed one half at a time: `bf4ebd2` narrowed *which* path counts as correct, leaving the arm still a whole-file positive `test()`. SC01 now has a negative arm too.

I did not take the suggested predicate, because measuring it against the corpus showed it would have shipped false positives. `node\s+(?!\$\{CLAUDE_SKILL_DIR\})[\w./-]*scripts/` flags any interpreter-invoked path containing `scripts/`, and two legitimate forms match it in the live tree:

- a **repo-level** script — create-skill's own SKILL.md documents `bash scripts/sync-symlinks.sh`, which is not this skill's asset, so `${CLAUDE_SKILL_DIR}` would be the wrong path for it;
- a **documentation placeholder** — `playwright-trace-analyzer` writes `node <skill_dir>/scripts/trace-extract.mjs`, a stand-in for the variable rather than a cwd-relative path.

So the arm is scoped to a path routing through the skill's *own* `scripts/` directory, identified by the skill directory's name, and requires an interpreter prefix so a prose reference or a markdown link by repo-relative path stays legal — the rule governs invocation, not mention. Measured across all four corpus skills that ship a `scripts/` directory: zero new findings, corpus unchanged at 24 clean / 31 failing of 55. The three remaining SC01 warnings all come from the pre-existing positive arm.

Two self-test fixtures added, since the three existing ones all vary only whether a correct path is present and so cannot reach this gap: a body with both a correct and an own-path cwd-relative invocation must fire, and one with a repo-level invocation must not. Both mutation-proven — disabling the arm reds the first alone, dropping the own-path scoping reds the second. Self-test 23 → 25 assertions.

---

skills/authoring/create-skill/SKILL.md
```diff
@@ -195,35 +212,39 @@ If the user is publishing the skill via `npx skills add` only, skip steps
...
+before declaring the skill done. Report the validator's own `Self-check:` line
+verbatim — it prints its own pass/total, so never restate that count from
+memory or invent one — then confirm the remaining `(judgment)` items passed.
+On failure:
 
-or, on failure:
-
-```
+```text
 Self-check: FAIL — fix these:
```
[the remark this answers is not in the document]
mthines: Addressed in 9dd20b4. ✅

Correct — `Self-check: FAIL — fix these:` is a form `printHuman` never emits.

One correction to the suggested replacement, though: it keeps the banner first and a `- [ ]` checklist line under it, and that ordering is also not what the validator prints. `printHuman` emits each finding *first*, as `<LEVEL> <ID> <file>:<line> — <message>`, and the `Self-check:` banner last. So I captured the output from an actual failing run rather than composing it:

```text
FAIL FM05 SKILL.md:3 — description must be non-empty and <= 1024 chars, got 1180
Self-check: FAIL — 1 failing
```

Kept internally consistent at one FAIL line and a count of 1 — worth being exact here, since the surrounding paragraph is the one telling readers to quote that line verbatim.

---

scripts/eval/l1.mjs
```diff
@@ -6335,4 +6335,104 @@ const isPollBlock = (block) =>
...
+  s.check("G51e SKILL.md has no cwd-relative validate-skill.mjs invocation",
+    cwdRelative.length === 0,
+    cwdRelative.length ? `${cwdRelative.length} cwd-relative invocation(s): ${cwdRelative.join("; ")}` : "");
+  const validatorSrc = existsSync(VALIDATOR) ? readFileSync(VALIDATOR, "utf8") : "";
+  s.check("G51e the validator's --self-test covers SC01 firing on a bare-variable prose mention",
+    /\$\{CLAUDE_SKILL_DIR\}`-anchored[\s\S]{0,400}?prose mention only: SC01 fires/.test(validatorSrc),
```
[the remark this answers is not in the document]
mthines: Addressed in 7887e0f. ✅

Agreed, and this was already on my own list from the iteration-2 simplify pass — an untraceable `[\s\S]{0,400}` window in a script whose header rule requires every constant to name its source. It was wrong in both directions: rewording the fixture reds L1 while the coverage is intact (your receipt), and moving the coverage elsewhere greens it while the behaviour is gone.

Replaced with three behavioural probes that build a fixture skill and run the validator against it, so the guard survives any rewording and reds only when SC01 itself stops discriminating:

- fires on a bare-variable prose mention with every command cwd-relative
- fires when a correct path coexists with an own-path cwd-relative one
- stays silent on a correct path alone

The silent probe is not padding — without it, an SC01 hardwired to fire would satisfy both firing rows. All three mutation-proven: rewording the fixture now keeps L1 green at 1636/1636; disabling the negative arm reds the coexist probe; hardwiring SC01 to fire reds the silent one. In each failing case `G51a` reds alongside `G51e`, the two guards being deliberately independent.

L1 1634 → 1636 (one row replaced by three).

---

scripts/eval/l1.mjs
```diff
@@ -6335,4 +6335,140 @@ const isPollBlock = (block) =>
...
+      ? "no ${CLAUDE_SKILL_DIR}/scripts/validate-skill.mjs invocation in SKILL.md"
+      : "SKILL.md not found");
+  // The convention is not SKILL.md's alone. The blocking finding this guard answers named
+  // four sites across three files, so an arm reading only `skillMd` enforces the rule on one
+  // of them and leaves the other three free to regress — reverting rules/quality-checklist.md
+  // to the cwd-relative form left this file green. Walk every markdown file the skill ships.
+  const cwdRelative = [];
+  for (const path of walk(CS, ".md")) {
+    const body = readFileSync(path, "utf8");
+    for (const hit of body.match(/(?:node|npx|bash|sh)\s+(?:[^\s"'`]*\/)?create-skill\/scripts\/[^\s"'`)]+/g) ?? []) {
```
mthines: issue (medium): 🟡 **Guard misses the bare cwd-relative form** **(non-blocking)**

The pattern requires a `create-skill/` path segment, so the shorter form the check's own name forbids passes. Rewriting the same site as `node scripts/validate-skill.mjs` leaves L1 green.

Evidence: `skills/authoring/create-skill/rules/quality-checklist.md:41` (bare form, L1 stayed green) · `skills/authoring/create-skill/rules/scripts-and-assets.md:36` (convention, no shape carve-out)

<sup>`pr-reviewer` · commit `94818ae`</sup>



---

skills/authoring/create-skill/rules/scripts-and-assets.md
```diff
@@ -0,0 +1,107 @@
...
+A script that mutates many files, or that cannot be trivially undone,
+runs in three passes rather than one:
+
+1. **Plan** — compute what would change; print a summary.
+2. **Validate** — check the plan against invariants (no unintended
+   deletions, counts match expectations).
+3. **Execute** — apply, only after validation passes.
+
+```bash
+node scripts/migrate.py --plan-only   # prints the diff, changes nothing
```
mthines: issue (low): ⚪ **Example breaks two of this file's own rules** **(non-blocking)**

The only cwd-relative script paths in the file that states the `${CLAUDE_SKILL_DIR}` convention 50 lines above, and `node` cannot execute a `.py` file either way.

Evidence: `skills/authoring/create-skill/rules/scripts-and-assets.md:43` (the convention, same file) · `skills/authoring/create-skill/rules/scripts-and-assets.md:64` (its own correct example)

```bash
python3 ${CLAUDE_SKILL_DIR}/scripts/migrate.py --plan-only
python3 ${CLAUDE_SKILL_DIR}/scripts/migrate.py --verify
python3 ${CLAUDE_SKILL_DIR}/scripts/migrate.py --apply
```

<sup>`pr-reviewer` · commit `94818ae`</sup>



---

skills/authoring/create-skill/scripts/validate-skill.mjs
```diff
@@ -0,0 +1,863 @@
...
+    //   - a documentation placeholder (`node <skill_dir>/scripts/x.mjs`) — a stand-in for the
+    //     variable, not a cwd-relative path.
+    // Both occur in the live corpus (create-skill and playwright-trace-analyzer respectively);
+    // measured across all four skills that ship a scripts/ directory, this predicate is clean.
+    //
+    // An interpreter prefix is required so a prose reference or a markdown link to the script
+    // by repo-relative path stays legal — the rule governs how a script is INVOKED.
+    const ownScripts = basename(dir).replace(/[.*+?^${}()|[\]\\]/g, "\\$&") + "\\/scripts\\/";
+    const invocation = new RegExp(
+      `(?:node|npx|python3?|bash|sh|deno|bun|tsx)\\s+(?:[^\\s"'\`]*\\/)?${ownScripts}[^\\s"'\`)]+`, "g");
```
mthines: suggestion (low): ⚪ **Two copies of one predicate, already diverged** **(non-blocking)**

L1's G51e arm lists four interpreters against this nine, so `python3`/`bun`/`tsx` are caught here but not there. `deno` cannot match either list — the path follows `deno run`.

Evidence: `scripts/eval/l1.mjs:6433` (node|npx|bash|sh only)

<sup>`pr-reviewer` · commit `94818ae`</sup>



---

skills/authoring/create-skill/scripts/validate-skill.mjs
```diff
@@ -0,0 +1,863 @@
...
+    //   - a repo-level script (`bash scripts/sync-symlinks.sh`) — not this skill's asset, so
+    //     ${CLAUDE_SKILL_DIR} would be the wrong path for it;
+    //   - a documentation placeholder (`node <skill_dir>/scripts/x.mjs`) — a stand-in for the
+    //     variable, not a cwd-relative path.
+    // Both occur in the live corpus (create-skill and playwright-trace-analyzer respectively);
+    // measured across all four skills that ship a scripts/ directory, this predicate is clean.
+    //
+    // An interpreter prefix is required so a prose reference or a markdown link to the script
+    // by repo-relative path stays legal — the rule governs how a script is INVOKED.
+    const ownScripts = basename(dir).replace(/[.*+?^${}()|[\]\\]/g, "\\$&") + "\\/scripts\\/";
```
mthines: nitpick (low): `ctx.dirName` is already `basename(resolve(dir))` — reuse it rather than re-deriving it unresolved, as this function already does for FM04. **(non-blocking)**

<sup>`pr-reviewer` · commit `94818ae`</sup>

---

mthines (commented):
---

scripts/eval/l1.mjs
mthines: Addressed in 5b3a005. ✅

Reproduced your receipt first — rewriting `rules/quality-checklist.md:41` to `node scripts/validate-skill.mjs` did leave L1 green at 1636/1636. This is the same partial-coverage shape as the SKILL.md-only arm it replaced two commits earlier, one level down: the arm was keyed on a path *segment*, and the shorter form simply has no segment to match.

It now keys on the skill's own script **filenames**, read off `scripts/` at run time. Two properties fall out of reading the directory rather than hand-writing a pattern:

- the bare, `./`-prefixed and full repo-relative forms are all caught, since none of them can avoid naming the file they run;
- a repo-level script is excluded by construction, with no allowlist to maintain — `bash scripts/sync-symlinks.sh` at `SKILL.md:200` is not a file in this skill's `scripts/`, so it stays legal, which is right because `${CLAUDE_SKILL_DIR}` would be the wrong path for it.

Mutation-proven both ways: the bare form now reds L1 at 1637/1638 naming the file and the invocation, and the full repo-relative form still does.

One failure mode that keying on a directory listing introduces, so it is guarded too: if `scripts/` were emptied or renamed, the arm would silently match nothing and pass. A second row now asserts the directory actually holds a script. L1 1637 → 1638.

---

skills/authoring/create-skill/rules/scripts-and-assets.md
mthines: Addressed in 26f1668. ✅

Both halves confirmed — the paths were the form the section forbids 50 lines above, and `node` cannot execute a `.py` file either way. Applied your suggested block verbatim (`python3 ${CLAUDE_SKILL_DIR}/scripts/migrate.py`), then swept the rest of the file for the same shape; no others remain.

One thing I deliberately did **not** do: extend `G51e` to catch this. `migrate.py` is a hypothetical example script, not a file in the skill's own `scripts/` directory, and the mechanical guard keys on real filenames precisely so it cannot start flagging illustrative paths it has no way to verify. So this class stays a review find rather than becoming a check — worth stating explicitly, since the rest of this PR has been converting exactly such finds into guards.

---

skills/authoring/create-skill/scripts/validate-skill.mjs
mthines: Addressed in 585b6db. ✅

Both parts confirmed, and the `deno` half is the sharper find: the real form is `deno run x.ts`, so `run` lands in the path slot and a bare `deno` alternative can never match anything. It was dead in both copies.

Unified into one `INTERPRETER` alternation, defined in the validator and copied byte-identically into `G51e`, with `run` optional for `deno` and `bun`, and a leading `\b` so `mysh` no longer matches via `sh` (your other observation — you dropped it as permissive-only, which is right, but it was free to fix here).

On sharing rather than copying: I kept the copy, deliberately. `G51e` exists to hold the validator to a contract from *outside* it, and importing the alternation would let a change to the validator silently change the guard meant to catch that change. But since the cost of a copy is exactly the drift you found, the two are now **asserted equal** rather than trusted to stay in step — mutation-proven by adding one interpreter to the validator's list, which reds that assertion by name. L1 1636 → 1637.

A self-test fixture pins the `deno run` form; reverting to the bare `deno|bun` alternative reds it alone. Self-test 25 → 26 assertions.

---

skills/authoring/create-skill/scripts/validate-skill.mjs
mthines: Addressed in 13458d9. ✅

Now uses `ctx.dirName`, same as FM04 does.

Worth recording that your verifier was right to keep this a nitpick rather than promoting it to a bug — I probed the tempting reading before writing the fix, and it does not hold. `basename(".")` is `"."` while `basename(resolve("."))` is the real directory name, which looks like a live defect, but `main()` resolves the positional argument at `validate-skill.mjs:847` before `validateSkill` ever sees it, so on the CLI path `dir` is always absolute and the two expressions agree. I ran the relative-`.` case end to end and SC01 behaved identically both ways.

It is still worth using the existing value, for a narrower reason than "a bug": `validateSkill` is exported, so a library caller can pass a relative path, and there the two genuinely diverge. `dirName` is the one that stays correct.

---

skills/authoring/create-skill/scripts/validate-skill.mjs
```diff
@@ -0,0 +1,885 @@
...
+    //   - a repo-level script (`bash scripts/sync-symlinks.sh`) — not this skill's asset, so
+    //     ${CLAUDE_SKILL_DIR} would be the wrong path for it;
+    //   - a documentation placeholder (`node <skill_dir>/scripts/x.mjs`) — a stand-in for the
+    //     variable, not a cwd-relative path.
+    // Both occur in the live corpus (create-skill and playwright-trace-analyzer respectively);
+    // measured across all four skills that ship a scripts/ directory, this predicate is clean.
+    //
+    // An interpreter prefix is required so a prose reference or a markdown link to the script
+    // by repo-relative path stays legal — the rule governs how a script is INVOKED.
+    const ownScripts = dirName.replace(/[.*+?^${}()|[\]\\]/g, "\\$&") + "\\/scripts\\/";
```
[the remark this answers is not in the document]
mthines: Addressed in d89c7aa. ✅

Confirmed on a fixture before fixing: SC01 was silent on `node scripts/check.mjs` while `G51e` reds on the same line. Two guards disagreeing about one rule, with no shape carve-out in `rules/scripts-and-assets.md:61` to justify the split.

SC01 now has two arms, because one prefix shape genuinely cannot express both cases — the `dirName` arm catches any depth of repo-relative path, and a second arm catches the bare and `./`-prefixed forms, which have no directory segment to key on at all. The second is keyed on the skill's own script **filenames**, which is exactly what keeps both deliberately-protected forms legal: `sync-symlinks.sh` is not a file in create-skill's `scripts/`, and a `<skill_dir>/` placeholder is neither bare nor `./`-prefixed.

Thank you for measuring the wholesale port and reporting the refutation rather than just the proposal — that saved a false-positive regression, since it fires on four `node <skill_dir>/scripts/*.mjs` placeholders in `playwright-trace-analyzer` that SC01's own comment exists to protect. I re-measured the narrower arm the same way: zero new findings across all 55 skills, both protected forms verified silent by name, corpus unchanged at 24 clean / 31 failing.

Fixture added for both the bare and `./` forms; disabling the arm reds it alone. Self-test 26 → 27 assertions.

---

scripts/eval/l1.mjs
```diff
@@ -6335,4 +6335,170 @@ const isPollBlock = (block) =>
...
+  //
+  // Keyed on the skill's OWN script FILENAMES, not on a `create-skill/` path segment. The
+  // segment form let the bare `node scripts/validate-skill.mjs` through untouched — the same
+  // partial-coverage shape as the SKILL.md-only arm it replaced, one level down. Reading the
+  // filenames off the directory also excludes a repo-level script by construction:
+  // `bash scripts/sync-symlinks.sh`, which SKILL.md legitimately documents, is not a file in
+  // this skill's scripts/ directory, so no allowlist has to be maintained by hand.
+  const ownScriptsDir = join(CS, "scripts");
+  const ownScriptNames = existsSync(ownScriptsDir)
+    ? readdirSync(ownScriptsDir).filter((n) => /\.(mjs|cjs|js|ts|py|sh)$/.test(n)).sort()
```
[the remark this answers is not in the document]
mthines: Partially addressed in 469544b — one list removed, the other declined with the reason recorded in code. ✅

I reproduced each gap independently, isolating one variable at a time rather than relying on your combined `.rb` + `ruby` receipt: `node scripts/helper.rb` (interpreter listed, extension not) left L1 green at 1638/1638, and so did `ruby scripts/helper.sh` (extension listed, interpreter not). Both lists really are load-bearing.

**The extension filter is gone**, fixed by deletion rather than by another assertion: the arm now takes every file in `scripts/` as a candidate name, so it cannot fall behind when the directory grows. Receipt A now reds at 1637/1638 naming the file. A non-script file landing in the list is harmless — nothing invokes a README through an interpreter.

**The `INTERPRETER` alternation I am deliberately not fixing**, and I want to be explicit rather than quietly closing the thread: `ruby scripts/x.sh` stays uncaught, and receipt B stays green. No enumeration of interpreters is complete, so an assertion saying "this list is complete" would be a check that cannot establish what it claims — the same objection that retired `G51e`'s prose regex earlier in this PR. The limitation is now written into the code with the reason it is the safe direction: a missing interpreter is a missed violation, never a false accusation, and this repo's skills invoke scripts with node, bash and python only.

On your broader observation — this being the fourth confirmed finding on `G51e` across three SHAs, each iteration narrowing to kill a false positive and reopening a false negative on the other side — I think that is the right read, and it is why the second half above is a documented boundary rather than a fifth narrowing. Worth noting the two sides are no longer symmetric: false positives here are silent and land on other people's skills, while false negatives are confined to create-skill's own docs and are visible to the next reviewer. So stopping on the false-negative side is the cheaper place to stop.

---

